PR #6443, opened 2023-08-02
bug/4702-Dylan-InlineErrorSelectAddress

Sparowhawk:
## Description of Change
Added inline error if user clicks Use this address and didn't select something

## Screenshots/Video
Toggle: <details><summary>Before/after:</summary>&nbsp;&nbsp;</details>

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Yes, so this should be treated like any other form error...with an inline error. Could we do something like what we have for gender ID? Don't mind the crappy screenshot; it's from an old design file (this is when user has tapped save without making a selection, which is what's happening in this bug, too). Just to give an idea of what I'm talking about. Copy should be "Select an address"
Screenshot 2023-08-02 at 1 35 52 PM

## PR Checklist

## For QA


theodur (approved):

alexandec:
I'd recommend one more change which is to make the error message go away when the user taps an address. They've fixed the problem when they tap, and that way they'll get immediate feedback that it's fixed

alexandec (approved):

Sparowhawk:
> I'd recommend one more change which is to make the error message go away when the user taps an address. They've fixed the problem when they tap, and that way they'll get immediate feedback that it's fixed. Feel free to ignore if our new approach is to not modify the error message until they try to submit again

elsewhere in the app and in this section in particular it doesn't update until they hit save. So going to keep that consistent until told otherwise.

alexandec:
Maybe it just depends which part of the app, haha, I've noticed both approaches in different parts. I think for a11y it makes sense to not change the message until they try to submit though, to avoid screen reader confusion

PR #6456, opened 2023-08-07
5493 claims letters maintenance window

kpethtel:
***DRAFT***

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/components/CommonErrorComponents/DowntimeError.tsx
```diff
@@ -31,9 +30,12 @@ const DowntimeError: FC<DowntimeErrorProps> = ({ screenID, overrideFeatureName }
...
   }
   const { downtimeWindowsByFeature } = useSelector<RootState, ErrorsState>((state) => state.errors)
-  const feature = DowntimeScreenIDToFeature[screenID]
-  const featureName = overrideFeatureName ? overrideFeatureName : downtimeWindowsByFeature[feature]?.featureName
-  const endTime = downtimeWindowsByFeature[feature]?.endTime.toFormat('fff')
+  const feature = ScreenIDToDowntimeFeature[screenID]
+  const downtimeWindow = feature ? downtimeWindowsByFeature[feature] : null
+
+  const featureName = ScreenNameFeatureConstants[screenID]
+  // const featureName = t(`downtime.${screenID}.featureName`)
```
kpethtel, line 37: Because the feature names are only used for copy, I tried moving them into the translations file but this line caused the linting error: `No overload matches this call.` I assume this means that typescript can't be certain that this will always return a value.

VAMobile/src/components/CommonErrorComponents/DowntimeError.tsx
```diff
@@ -31,9 +30,12 @@ const DowntimeError: FC<DowntimeErrorProps> = ({ screenID, overrideFeatureName }
     mb: theme.dimensions.contentMarginBottom,
   }
   const { downtimeWindowsByFeature } = useSelector<RootState, ErrorsState>((state) => state.errors)
-  const feature = DowntimeScreenIDToFeature[screenID]
-  const featureName = overrideFeatureName ? overrideFeatureName : downtimeWindowsByFeature[feature]?.featureName
-  const endTime = downtimeWindowsByFeature[feature]?.endTime.toFormat('fff')
+  const feature = ScreenIDToDowntimeFeature[screenID]
+  const downtimeWindow = feature ? downtimeWindowsByFeature[feature] : null
```
kpethtel, line 34: This type of pattern was a recurring pain point I experienced with TS. I would expect that there's probably a better pattern for this.

VAMobile/src/components/CommonErrorComponents/DowntimeError.tsx
```diff
@@ -3,22 +3,21 @@ import { useTranslation } from 'react-i18next'
...
+import { ScreenIDToDowntimeFeature, ScreenIDTypes, ScreenNameFeatureConstants } from 'store/api/types'
 import { useSelector } from 'react-redux'
 import { useTheme } from 'utils/hooks'
 
+
 export type DowntimeErrorProps = {
   /**The screen id for the screen that has the errors*/
   screenID: ScreenIDTypes
-  /**Override the feature name in the event that a feature happens to share the same api error(ex:contact information and personal information) */
-  overrideFeatureName?: string
```
kpethtel, line 17: Removing this because I'm changing the pattern to be driven by screen id, which *should* be unique, which I believe solves the problem this was intended to fix. In theory. Plausible we could want more complex logic in the future.

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/ContactInformationScreen.tsx
```diff
@@ -89,7 +89,7 @@ const ContactInformationScreen: FC<ContactInformationScreenProps> = ({ navigatio
   useFocusEffect(
     React.useCallback(() => {
       if (needsDataLoad && profileNotInDowntime) {
-        dispatch(getProfileInfo(ScreenIDTypesConstants.PERSONAL_INFORMATION_SCREEN_ID))
+        dispatch(getProfileInfo(ScreenIDTypesConstants.CONTACT_INFORMATION_SCREEN_ID))
```
kpethtel, line 92: Changing this because I'm making it more screen id driven and this felt like it was given the wrong screen id in order to use another screen's values.

VAMobile/src/components/CommonErrorComponents/DowntimeError.tsx
```diff
@@ -31,9 +29,12 @@ const DowntimeError: FC<DowntimeErrorProps> = ({ screenID, overrideFeatureName }
...
   const { downtimeWindowsByFeature } = useSelector<RootState, ErrorsState>((state) => state.errors)
-  const feature = DowntimeScreenIDToFeature[screenID]
-  const featureName = overrideFeatureName ? overrideFeatureName : downtimeWindowsByFeature[feature]?.featureName
-  const endTime = downtimeWindowsByFeature[feature]?.endTime.toFormat('fff')
+  const feature = ScreenIDToDowntimeFeature[screenID]
+  const downtimeWindow = feature ? downtimeWindowsByFeature[feature] : null
+
+  const featureName = ScreenNameFeatureConstants[screenID]
+  // const featureName = t(`downtime.${screenID}.featureName`)
+  const endTime = downtimeWindow?.endTime.toFormat('fff')
```
kpethtel, line 37: This is one of the reasons I didn't bother making `ScreenIDToDowntimeFeature` 1:many. If we did that, we'd need to define rules for how it's handled when multiple are down. Which endtime does it use?

VAMobile/src/store/slices/errorSlice.ts
```diff
@@ -88,6 +79,15 @@ export const checkForDowntimeErrors = (): AppThunk => async (dispatch) => {
       [maintWindow.service]: metadata,
     }
   }
+  // const metadata: DowntimeWindow = {
```
kpethtel, line 82: This is just testing code. Ignore.

VAMobile/src/translations/en/common.json
```diff
@@ -58,6 +58,18 @@
   "done": "Done",
   "done.a11yHint": "Exits out of the web view and navigates to the previous page",
   "done.picker.a11yHint": "confirms selection and closes the menu",
+  "downtime.CLAIMS_HISTORY_SCREEN.featureName": "Claims",
```
kpethtel, line 61: This was from my attempt to make the labels translations. Wouldn't be in the final product unless we decided to make it work correctly.

VAMobile/src/store/slices/errorSlice.ts
```diff
@@ -24,7 +17,6 @@ export type DowntimeWindowsByFeatureType = {
 }
 
 export type DowntimeWindow = {
-  featureName: string
```
kpethtel, line 27: Removed this because it should be determined by screen, not by upstream feature name.

VAMobile/src/utils/hooks/index.tsx
```diff
@@ -32,17 +32,29 @@ import { useTheme as styledComponentsUseTheme } from 'styled-components'
...
   const { downtimeWindowsByFeature } = useSelector<RootState, ErrorsState>((state) => state.errors)
-  const mw = downtimeWindowsByFeature[feature]
-  if (!!mw && mw.startTime <= DateTime.now() && DateTime.now() <= mw.endTime) {
-    return true
-  }
-  return false
+  return featureInDowntime(feature, downtimeWindowsByFeature)
+}
+
+// this is not a hook. not sure this is the right place for this function
```
kpethtel, line 50: These two are not actual hooks, but I didn't know where else to put them. Also, I found that I had to write them to receive `downtimeWindowsByFeature` rather than fetch them from state due to a linting issue saying you can't use hooks in callbacks. Very likely there's a better way of doing this.

VAMobile/src/store/api/types/Errors.ts
```diff
@@ -79,43 +78,33 @@ export const DowntimeFeatureTypeConstants: {
...
-  [DowntimeFeatureTypeConstants.secureMessaging]: 'Secure Messaging',
-  [DowntimeFeatureTypeConstants.letters]: 'Letters',
-  [DowntimeFeatureTypeConstants.directDepositBenefits]: 'Direct Deposit',
-  [DowntimeFeatureTypeConstants.disabilityRating]: 'Disability Rating',
-  [DowntimeFeatureTypeConstants.militaryServiceHistory]: 'Military Service History',
-  [DowntimeFeatureTypeConstants.appeals]: 'Appeals',
-  [DowntimeFeatureTypeConstants.userProfileUpdate]: 'Personal Information',
-  [DowntimeFeatureTypeConstants.payments]: 'Payments',
-  [DowntimeFeatureTypeConstants.rx]: 'VA Prescriptions',
+export const ScreenIDToFeatureName = {
```
kpethtel, line 81: I've changed this to be more screen driven and less feature driven. To me, it feels more natural to combine pieces of data to make the relationships clearer, like:
```
screenId: {
  name: 'what i want the user to see',
  features: ['claims', 'appeals', 'whatever']
}
```

But I'm new to typescript and chose not to change too much. Let me know if anyone else would find that clearer and I could give it a try.

dumathane (approved):

Sparowhawk (approved):

alexandec (approved):

dumathane (approved):

Sparowhawk (dismissed):

VAMobile/src/utils/hooks/index.tsx
```diff
@@ -35,16 +35,32 @@ import { useTheme as styledComponentsUseTheme } from 'styled-components'
...
+  return oneOfFeaturesInDowntime(features, downtimeWindowsByFeature)
+}
+
+export const featureInDowntime = (feature: DowntimeFeatureType, downtimeWindows: DowntimeWindowsByFeatureType): boolean => {
+  const mw = downtimeWindows[feature]
+  return !!mw && mw.startTime <= DateTime.now() && DateTime.now() <= mw.endTime
+}
+
+export const oneOfFeaturesInDowntime = (features: DowntimeFeatureType[], downtimeWindows: DowntimeWindowsByFeatureType): boolean => {
+  return !!features?.some((feature) => featureInDowntime(feature as DowntimeFeatureType, downtimeWindows))
```
theodur, line 63: I don't feel too strongly about this, but since this file is intended for hooks, and there are multiple helper functions for downtime, it might be worth making a separate file for downtime utils in the `VAMobile/src/utils` directory and moving these there

theodur (approved):

VAMobile/src/utils/hooks/index.tsx
kpethtel: I agree. I don't want to hold this up any further but have a followup ticket for this and will move these methods at that time.

DJUltraTom (approved):

dumathane (approved):

theodur (approved):

PR #6467, opened 2023-08-08
bug/4985-theo-DisabilityRatingsAPICall

theodur:
## Description of Change

~~Disability ratings data is currently fetched on the sync screen before the home screen is opened. Since we no longer display disability ratings on the home page, we don't need the API call in the sync screen. This PR removes the call from the sync screen and adds it to the `BenefitsScreen` instead, since it's needed for the `Disability rating` nav button. I also added a screen ID for the benefits screen, since it doesn't have one.~~

Turns out we'll still need the disability rating API call on the sync screen for the upcoming veteran status card. The sync screen currently makes duplicate API calls because there are multiple dependencies in the `useEffect` hook for loading data. When 1 API call finishes first, and its `preloadComplete` state is set to `true`, the code in the effect is re-triggered, and based on the current logic, it'll refetch data if the `preloadComplete` state for the other slices isn't set to `true`. I updated the logic to prevent duplicate API calls from being made if the data for the other slices is still being loaded.

## Screenshots/Video

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/9042882/a4937107-e043-4be2-adcb-b15221cd9d84

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

- [ ] Verify home screen loads as normal after signing in
- [ ] Verify the benefits screen loads the ratings data for the nav button

## PR Checklist

## For QA


Sparowhawk (dismissed):

dumathane:
I think this might mess up Veteran Status...chatting with Theo.

dumathane (changes requested):
I'd like to hold on this until a VeteranStatus decision is made. I will need disability rating on the sync screen if we can.

VAMobile/src/screens/SyncScreen/SyncScreen.tsx
```diff
@@ -47,15 +47,26 @@ const SyncScreen: FC<SyncScreenProps> = () => {
...
+    loggedIn,
+    personalInformationLoaded,
+    personalInformationLoading,
+    militaryInfoAuthorization,
+    authorizedServicesLoaded,
+    disabilityRatingLoaded,
+    disabilityRatingLoading,
+    militaryHistoryLoaded,
+    militaryHistoryLoading,
+  ])
```
theodur, line 69: Based on how beefy this dependency array is getting, this could be separated into multiple `useEffect`s. I kept it in one since these functions all depend on the state of `loggedIn`, but if anyone feels strongly about this I can split it out.

theodur:
@dumathane I added the API call back to the sync screen since it'll still be needed. Instead, I resolved the issue with the duplicate API calls

Sparowhawk (approved):

VAMobile/src/screens/SyncScreen/SyncScreen.tsx
alexandec: My guess is this is fine, since even though it will run multiple times, it should be quick to run. I would think if you don't see any noticeable performance degradation, it's fine. It's kind of strange that it was written this way to begin with though, not sure why.

alexandec (approved):

VAMobile/src/screens/SyncScreen/SyncScreen.tsx
theodur: Yeah I agree, it's interesting how this was written. My guess is to consolidate the `useEffect` calls, although it's debatable whether structuring it like this has any strong advantages. It means this will be called more frequently, but shouldn't be an issue since it's relatively quick like you mentioned. FWIW, the `getProfileInfo` API call is going to be split into smaller separate calls in the near future, so this will almost certainly need to be refactored anyway

dumathane (approved):

PR #6478, opened 2023-08-11
UI Automation/3709-Chika-detox-ui-test-sign-in

cadibemma:
## Description of Change

Detox test for Sign In to align with RC test steps

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

- [ ] Create detox test that covers all cases for this section

## PR Checklist

## For QA


timwright12:
I'm getting consistant fails on iOS for both tests:
```
Test Failed: Timed out while waiting for expectation: TOEXIST WITH MATCHER(id == “Login-page”) TIMEOUT(12s)
```

and

```
Test Failed: No elements found for “MATCHER(text == “Back”)”
```

Feel free to ignore me if everyone else is passing though

theodur (approved):

VAMobile/e2e/tests/SignIn.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+    'We’ll verify your identity through a secure process from ID.me or Login.gov. This trusted partner provides the strongest identity verification system available.',
+  LOA_GATE_READ_MORE_P2: 'To complete the process on your smartphone, you’ll need these items:',
+  CONTINUE_SIGN_IN_BTN_ID: 'Continue to sign in',
+}
+
+describe('Sign In', () => {
+  it('should show sign in page content', async () => {
+    await waitFor(element(by.id(SignE2eIdConstants.LOGIN_PAGE_ID)))
+      .toExist()
+      .withTimeout(2000)
```
alexandec, line 20: This 2-second timeout can cause the test to fail if someone's machine is slow, or when running on CI. I found this failure happening locally for me some of the time when running these tests

alexandec:
Passes for me on Android and iOS, however I added a note above about a timeout.

Another question, how do these tests relate to the tests in `LoginScreen.e2e.ts`?

VAMobile/e2e/tests/SignIn.e2e.ts
theodur: That's a good point. I also wonder if we should create constants for our timeouts based on how fast/slow we need them to be. I noticed we have a timeout of `10000` in a few tests, so it might be worth consolidating that

cadibemma: Is that something that should changed in this PR or will another ticket be created for that change? @alexandec @theodur

Issue #6386, opened 2023-07-26
Create documentation for Segmented Control component


jessicawoodin:
### Description
Based on the approved component documentation template (see #6301), we need to update the documentation for the Segmented Control component. A draft has been started (see #4599).

### Steps
- [x] Update documentation in #4599 to match approved template (Holly may be able to answer any questions)
- [x] Update Figma design library
     - [x] Implement [plan](https://app.zenhub.com/workspaces/va-mobile-product-view-610035bc5395bb000e62e529/issues/gh/department-of-veterans-affairs/va-mobile-app/5457) to simplify the Assets panel by removing frames/groups and nesting/hiding lower level components
     - [x] Implement [component template](https://app.zenhub.com/workspaces/va-mobile-product-view-610035bc5395bb000e62e529/issues/gh/department-of-veterans-affairs/va-mobile-app/5457)
     - [x] Update metadata with description, link and alternative names
- [x] Review with UX team
     - [x] Share accessibility draft with Brea
     - [x] Share content draft with Misty
     - [x] Share draft with Holly
     - [x] Share draft with Jen
- [x] Share draft with platform team
     - [x] Add iframe for [Storybook](https://department-of-veterans-affairs.github.io/va-mobile-temporary/iframe.html?viewMode=story&id=button--basic)
- [x] Publish documentation to doc site
     - [x] Add link to component in Figma design library
- [x] Close #4599 and update [component documentation spreadsheet](https://docs.google.com/spreadsheets/d/1_EAH2LWSzwF8Om7o4LAYJf6gT9UWENANYpF7SZy3j8w/edit#gid=0)

Issue #6383, opened 2023-07-25
Update documentation site to incorporate design system


jessicawoodin:
### Description
As part of the platform work, we'll be adding both onboarding and component documentation for the design system to the doc site. As part of this ticket, we'll update the doc site to incorporate the design system according to the [MVP diagram in FigJam](https://www.figma.com/file/jsJy7ZGkxqHfLf08rJ7t5y/VA-Mobile-App-Platform-(Post-MVP)?type=whiteboard&node-id=48-9861&t=lflLnXcmffZhaQJh-4).

### Steps
- [x] Add "design system" section to doc site
- [x] Add subsections that align with [VA design system website](https://design.va.gov/)
- [x] Move relevant pages from "UX" and "front end" subsections to "design system" section ([see Taxonomy doc](https://docs.google.com/spreadsheets/d/1RQpy9tClEJULqV7hU9Rx8eK9fsnzHsSZe6KgA5diAVU/edit#gid=836335809))
- [x] Test that all pages were moved successfully

PR #6481, opened 2023-08-11
Adding design system

jessicawoodin:
## Description of Change

- Tickets #6383 and #6386
- Added design system to top nav and footer
- Created new sidebar for design system
- Moved pages from docs to design folder
- Created new placeholder pages for design system content
- Created new segmented control documentation

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


timwright12 (changes requested):
@jessicawoodin I think we need to move all the components that are not in the design system out of this section and back into the normal parts of the site. I don't think we've spoke about what will be a component vs. a pattern yet (cc @drjecker) we also don't know if the code will be the same after we migrate from the app into the system. I think we should leave all of it where it is and migrate one at a time as we get components moved over

VAMobile/documentation/design/Foundation/color-palette.md
```diff
@@ -0,0 +1,17 @@
...
+
+The VA mobile app has two color themes: Light and Dark. [The VA mobile app themes](https://www.figma.com/file/bGO6g5cCvWycrNjoK66PXc/VA-Mobile-Design-Tokens?node-id=151%3A76) use colors from the [VA style guide](https://design.va.gov/foundation/color-palette), and only deviate if necessary for accessibility purposes.
+
+<iframe width="550" height="500" src="https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FQVLPB3eOunmKrgQOuOt0SU%2F%25F0%259F%2593%2590-DesignLibrary2.0---VAMobile%3Ftype%3Ddesign%26node-id%3D3859%253A7446%26t%3DLWuS4oyNuplsuZBa-1" allowfullscreen></iframe>
+
+**Themes**
+- [Light theme in Figma](https://www.figma.com/file/yXL0MkEKyAPGXPZqRH0VFZ/VA-Mobile-light-theme?node-id=183%3A441)
+- [Dark theme in Figma](https://www.figma.com/file/gOhb2kZvoQiXiGigqWZhnx/VA-Mobile-dark-theme?node-id=183%3A441)
+
+## Front-end
```
timwright12, line 14: This is the app, not the design system. I don't think it will be accurate. We should move this back until we're ready

VAMobile/documentation/design/About/content-writers.md
```diff
@@ -0,0 +1,6 @@
+---
+title: For content writers
+sidebar_position: 3
+---
+
+Content development is in progress and will be published by the end of Q3.
```
timwright12, line 6: Can we add draft: true to the front matter on empty pages like this and make sure they aren't in the nav until they're ready? https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-pages

jessicawoodin:
@timwright12 Hmm that doesn't match up with what was discussed last sprint and what I shared with Ryan in our last touchpoint. We have a lot of usage guidelines for components that UX needs to reference while we're still building all of the components into Storybook. I'm worried it'll be confusing if some usage guidelines are in the "Documentation" section and some are in the "Design System" section of the doc site. I'm wondering if there's another way to address your concerns than to move everything back to where it was. Can we discuss? Do we need a separate call from standup?

narin (approved):

timwright12 (approved):

PR #6503, opened 2023-08-15
bug/6292-theo-SendMessageReplyDraft

theodur:
## Description of Change

There is an API error that happens occasionally when sending a new message:
```
attempted to use reply draft in send message
```

This error recently started happening on July 5th, after the 2.6.0 release. After doing some investigating, this is most likely due to the `draft_id` field being sent with the post data when it shouldn't be. I couldn't find any recent changes that would have caused this, so just to be safe, this PR resets the saved draft ID when the `ReplyMessage` and `StartMessage` screens initially load. This doesn't get to the root cause of why `draft_id` is being sent when it shouldn't be, but it should hopefully prevent it from being sent when it shouldn't.

## Screenshots/Video

N/A

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

Verify that saving a draft works correctly on the `StartNewMessage` and `ReplyMessage` screens.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HealthScreen/SecureMessaging/ReplyMessage/ReplyMessage.tsx
+++ b/VAMobile/src/screens/HealthScreen/SecureMessaging/ReplyMessage/ReplyMessage.tsx
@@ -31,6 +31,7 @@ import {
   SecureMessagingState,
   dispatchSetActionStart,
   getMessageSignature,
+  resetSaveDraftComplete,
   resetSendMessageComplete,
   resetSendMessageFailed,
   saveDraft,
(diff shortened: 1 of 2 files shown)


narin:
@theodur This looks good. Just want to confirm you tested that saving drafts still works properly on these screens?

theodur:
> @theodur This looks good. Just want to confirm you tested that saving drafts still works properly on these screens?

@narin Yeah, I tested it last night when I initially made the draft PR, but I just tested again right now to make sure. We should be good to go ✅

dumathane (approved):

PR #6563, opened 2023-08-22
feature/6408-theo-DemographicsEndpoint

theodur:
## Description of Change

Updates the app to use the new [/v0/user/demographics](https://department-of-veterans-affairs.github.io/va-mobile-app/api/#/paths/~1v0~1user~1demographics/get) endpoint for fetching preferred name and gender identity data. Since gender identity and preferred name is still coming from `/v1/user` endpoint, I updated the `dispatchFinishGetProfileInfo` reducer to no longer store that information in the `profile` object, and instead, that data is now being stored in the `demographics` object. I also updated all mock & test data to reflect the new structure change of the personal information state. I kept the `genderIdentity` and `preferredName` values in the `profile` object for now because the `UserDataProfile` type still holds that data, since that's still being returned from the `v1/user` endpoint. Once we switch over to the new `v2/user` endpoint, those can be removed from the `profile` object.

There are a few ways we can organize the state for demographics that I thought of when working on this:
1. Keep `genderIdentity` and `preferredName` in the `profile` object. Although, since the data will no longer being fetched with the other profile data, it didn't seem natural to keep it under `profile`.
2. Create a new slice for `demographics` and store `preferredName` and `genderIdentty` at the root level of that slice. I was considering going this route, because in one way it would help further isolate the demographics data along with the actions for updating the gender identity and preferred name state. It would also help separate the loading and loaded state of demographics and profile info, since we wouldn't need a separate loading state for each of those in the same slice. I didn't go this route right now because it feels like demographics still fits under personal information. It also might be confusing if bits of user information are separated into multiple slices. If the demographics data grows much larger than it is now, it might be worth looking into this approach.
3. Adding a new `demographics` object in the personal information slice for storing data that comes from the new endpoint. This is the route I went because it seems to separate the demographics data from the profile data nicely without the need to add a new slice.

I'm open to other ways of organizing the state also. I feel like option 2 could be a better long term way of organizing the data, I just don't know if it's too soon for that right now. Also for reference, the new [user endpoint](https://department-of-veterans-affairs.github.io/va-mobile-app/api/#/paths/~1v2~1user/get) that we'll eventually be switching to only returns a few pieces of user information since data from the current `/v1/user` endpoint is being split into multiple endpoints. That might be worth taking into account when deciding how we want to organize user information state in general.

## Screenshots/Video

N/A since there are no visual changes being made.

## Testing

Tested loading the home screen after login, opening the profile and personal information screens. Also tested updating preferred name & gender identity and verifying that the updates are reflected in other screens afterward.

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

- Verify that the home screen only shows the preferred name or first name in the greeting (there shouldn't be any visual changes between the name when the screen is being loaded).
- Verify the home screen, profile screen, personal information screen, gender identity screen, and preferred name screen all load and show demographics data.
- Verify updates to preferred name and gender identity are reflected on other screens after being updated by the user.
- Verify that the app still functions correctly when the `v0/user/demographics` endpoint is down. Only the personal information screen should show an error since preferred name or gender identity data can't be retrieved, but other screens in the app should still work. The home screen in this case should fall back to using the user's first name in the greeting message. (This is why the screen ID is set to the personal information screen when fetching demographics on the home screen)

## PR Checklist

## For QA


dumathane:
Converting to draft until we hear from theo about the potential coding work around.

PR #6583, opened 2023-08-24
Bug/6356 theo-AndroidActionSheet

theodur:
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/SelectFile/SelectFile.tsx
+++ b/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/SelectFile/SelectFile.tsx
@@ -10,7 +10,7 @@ import { Events } from 'constants/analytics'
 import { MAX_TOTAL_FILE_SIZE_IN_BYTES, isValidFileType } from 'utils/claims'
 import { NAMESPACE } from 'constants/namespaces'
 import { logAnalyticsEvent, logNonFatalErrorToFirebase } from 'utils/analytics'
-import { useRouteNavigation, useShowActionSheet, useTheme } from 'utils/hooks'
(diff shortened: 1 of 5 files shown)


alexandec (approved):
Looks good, just have two questions:
- Would it be possible to integrate this functionality into `useShowActionSheet` itself?
- Would Android users expect the back button to navigate back and close the ActionSheet, versus just having the back button do nothing? I'm not sure what the convention is for that type of thing on Android

narin (approved):

theodur:
@alexandec Great questions, originally I tried adding this functionality directly to `useShowActionSheet`, but the issue I ran into was determining when an option is selected and the actionsheet is dismissed. Adding the `useBeforeNavBackListener` hook and having that prevent navigating back worked fine, but there doesn't seem to be a way to know when the actionsheet is closed with that approach. Unfortunately, the library we're using for actionsheets doesn't have a built-in way of programmatically getting that information. There have been a few issues in the library's repo that people have opened, asking for something similar. Accessing the `ActionSheetProvider` ref didn't work either 🙁, so the screen would be stuck with the back navigation being disabled until re-rendered. Not sure if there's a way around that, but maybe there might be with more discovery.

And I'm actually not sure what the convention is for Android and navigating back with action sheets 🤔. That's a good point, because if the default behavior for native Android action sheets are to close when the back button is pressed, user's may expect that. I know for our destructive action sheet, we prevent navigating backwards for both iOS and Android on screens that use them, so maybe we want that alignment here as well? This PR at least ensures the user won't have the action sheet stuck on the screen, but I think the behavior of back navigation for Android action sheets is worth looking into at a larger scale, for both destructive and non-destructive actionsheets. Maybe in the [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/6269) for combining the action sheet hooks or a separate CU ticket

dumathane:
>...If you check say the logging out confirmation alert?
I checked the contact info stuff and android seems to respect an alert as a something you can't native back from.

PR #6616, opened 2023-08-30
Feature/6407 roettger implement segmented control component

TimRoe:
## Description of Change

Implemented the design system Segmented Control component into the VA mobile app (except in secure messaging, split to #6647 as a special scenario involving redux state setting value). This entailed reworks on the package side as well as some minor app-level config tweaks.

No functional behavior changes result from these changes.*

* Technically not totally true, if you look closely it responds slightly to light/dark mode changes faster than the rest of the app but this is negligible.

For documentation, [the corresponding design system repo PR is here](https://github.com/department-of-veterans-affairs/va-mobile-temporary/pull/14).

## Screenshots/Video

iOS:

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/8690976/3963ff7f-b894-44b6-b1fe-95297a319957

Android:

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/8690976/be16bf0f-ffc9-460f-bfd8-82fffdb19875

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

- [x] use new component in each place listed above in the app
   - Secure Messaging split to #6647 
- [x] fix any broken unit tests
- [x] ensure the component still works as expected
   - [x] light/dark mode functions, etc.
   - [x] accessibility is in line with how components functioned previously
   - [x] Check if a11y hint works as intended when removing the same tag from the app translations (ensuring it's using the components package i18n)

## PR Checklist

## For QA


VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.test.tsx
```diff
@@ -208,7 +209,7 @@ context('ClaimsHistoryScreen', () => {
       })
 
       expect(testInstance.findAllByType(AlertBox).length).toEqual(1)
-      expect(testInstance.findAllByType(TextView)[5].props.children).toEqual('Claims status is unavailable')
```
TimRoe, line 211: I know we were redoing brittle unit tests as we encountered issues, but this ticket has already had way too many hang-ups so just did the quick fix.

timwright12 (approved):
tested locally and in testFlight, seems to work well!

VAMobile/tsconfig.json
```diff
@@ -17,6 +17,7 @@
 		"allowSyntheticDefaultImports": true,
 		"esModuleInterop": true,
 		"resolveJsonModule": true,
+		"skipLibCheck": true,
```
Sparowhawk, line 20: why did this line need to be added?

Sparowhawk (approved):

VAMobile/tsconfig.json
TimRoe: Short answer: without it there was a litany of (52) TS linting errors from `yarn lint:ci` within node_modules. 

When I dug into it, it seemed the likely culprit was having duplicate type definitions within packages--many seemed to be related to the design system having different (more up-to-date) versions of the i18next and styled-components libraries, but there was also some react/react-native. At some point in the troubleshooting, we were using directly incompatible i18n functionality, but maybe that is no longer the case having sorted out other issues. I didn't dig fully into the react/react-native linting issues, but my assumption was they related to having to include the expo library or it'd crash because the design system package references it even if it can never be hit when running within the app, but had the benefit of allowing the same code with no modifications to run both within the design system repo and as published npm package.

On the whole, it didn't seem like we should really care about TS problems within node_modules if things still run as expected and we should still get linting failures if the issues were seeping into app-level code.

TimRoe:
@rbontrager FYI: Tim W [already created a build](https://dsva.slack.com/archives/C022SV53ASZ/p1694006723863699) that thus far is still current with dev. 

Also left the Needs Review tag as a reminder to FE folks to look at it as it's the first design system integration even though it already has the requisite 2 approvals.

theodur (dismissed):

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/AppealDetailsScreen/AppealDetailsScreen.tsx
```diff
@@ -110,13 +111,7 @@ const AppealDetailsScreen: FC<AppealDetailsScreenProps> = ({ navigation, route }
...
           <TextView variant="MobileBody">{t('appealDetails.submitted', { date: formattedSubmittedDate })}</TextView>
           <Box mt={theme.dimensions.standardMarginBetween}>
-            <SegmentedControl
-              values={controlValues}
-              titles={controlValues}
-              onChange={setSelectedTab}
-              selected={controlValues.indexOf(selectedTab)}
-              accessibilityHints={segmentedControlA11yHints}
-            />
+            <SegmentedControl labels={controlLabels} onChange={setSelectedTab} selected={controlLabels.indexOf(selectedTab)} labelsA11yHints={segmentedControlA11yHints} />
```
dumathane, line 114: Did our variable names change because ya'll made edits to the component itself in the library and we're now matching?

dumathane (approved):

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/AppealDetailsScreen/AppealDetailsScreen.tsx
TimRoe: Yeah, I combined titles and values props into one since I didn't see the utility in having one array to give the labels shown visually and another array that was only used internally around setting selected state.

That said, there's a chance that'll need to be reworked due to SM. I don't think it's inherently part of the problems I ran into and more on using redux, but also made me realize it may behave finnicky when things change (e.g. the unread count in the segment label). I'll add another note to the separate ticket to be on the lookout for that.

TimRoe:
Updated to address an issue Rachael had noted with Android screen reader not reading correctly due to lacking accessibilityLabel (baked into existing component with testIdProps which was removed in the design system component).

Additionally, in doing so I noticed a bug with my simplification--if the segment label changed it would become unselected.  While not a problem on this ticket since all labels are static, would've been a problem for Secure Messaging. Instead of reintroducing the duplication of the app Segmented Control of two usually identical arrays of strings, the component now directly uses the index of selected labels (already there) to determine which is selected.

Not directly relevant to this ticket, but [also carried out an under-the-hood improvement per the design system PR](https://github.com/department-of-veterans-affairs/va-mobile-temporary/pull/14#discussion_r1317813452) that allowed removal of pulling in the `expo` library within the package, removing a lot of node_module bloat.

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimDetailsScreen.tsx
```diff
@@ -120,12 +121,12 @@
...
           <TextView variant="MobileBody">{t('claimDetails.receivedOn', { date: formattedReceivedDate })}</TextView>
           <Box mt={theme.dimensions.standardMarginBetween}>
-            <SegmentedControl values={controlValues} titles={controlValues} onChange={onTabChange} selected={controlValues.indexOf(selectedTab)} accessibilityHints={a11yHints} />
+            <SegmentedControl labels={controlLabels} onChange={onTabChange} selected={selectedTab} a11yHints={a11yHints} />
           </Box>
         </Box>
         <Box mt={theme.dimensions.condensedMarginBetween}>
-          {claim && selectedTab === t('claimDetails.status') && <ClaimStatus claim={claim || ({} as ClaimData)} claimType={claimType} />}
-          {claim && selectedTab === t('claimDetails.details') && <ClaimDetails claim={claim} />}
+          {claim && selectedTab === 0 && <ClaimStatus claim={claim || ({} as ClaimData)} claimType={claimType} />}
```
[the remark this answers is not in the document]
TimRoe: Neat, preexisting issue.

dumathane (approved):

Sparowhawk (approved):

PR #6620, opened 2023-08-30
Bug/binny 5659 missing from triage group bug

dumathane:
## Screenshots/Video

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/46909588/00a64499-a25c-4766-8d88-e434fee80f2b

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/store/slices/secureMessagingSlice.ts
```diff
@@ -519,7 +519,7 @@ export const sendMessage =
       if (isErrorObject(error)) {
         logNonFatalErrorToFirebase(error, `sendMessage: ${secureMessagingNonFatalErrorString}`)
         dispatch(dispatchFinishSendMessage(error))
-        showSnackBar(messages.errorMsg, dispatch, retryFunction, false, true)
+        replyToID && hasErrorCode(SecureMessagingErrorCodesConstants.TRIAGE_ERROR, error) ? '' : showSnackBar(messages.errorMsg, dispatch, retryFunction, false, true)
```
theodur, line 522: Since the empty string (`''`) isn't doing anything, the triage error logic could be stored in a variable, and you could do something like:

```ts
const isReplyWithTriageError = replyToID && hasErrorCode(SecureMessagingErrorCodesConstants.TRIAGE_ERROR, error)
!isReplyWithTriageError && showSnackBar(messages.errorMsg, dispatch, retryFunction, false, true)
```

VAMobile/src/components/MessageAlert.tsx
```diff
@@ -23,6 +28,16 @@ export type MessageAlertProps = {
 const MessageAlert: FC<MessageAlertProps> = ({ hasValidationError, saveDraftAttempted, scrollViewRef, focusOnError, errorList }) => {
   const theme = useTheme()
   const { t } = useTranslation(NAMESPACE.HEALTH)
+  const { replyTriageError } = useSelector<RootState, SecureMessagingState>((state) => state.secureMessaging)
```
theodur, line 31: At first glance, I felt like this and the related code could be removed from of the component, and a prop could be used instead for deciding when to display the triage error to keep the component simpler and more general (because this is exclusive to the reply message screen). But looking at what the error does, it'd be cumbersome to handle the resets when navigating back to the secure messaging screen if this and related code wasn't in here (although it could be done with `useBeforeNavBackListener` in the reply message screen), so I think this is fine for now. This component could get cluttered if we get more message errors types in the future, so if that happens, we may want to look into refactoring stuff like this

dumathane: I'd personally love to delete MessageAlert all together. It feels like a lot of layered on complexity for very little removal of duplicative code.

VAMobile/src/store/slices/secureMessagingSlice.ts
dumathane: Fixed as suggested.

Sparowhawk (dismissed):

VAMobile/src/components/MessageAlert.tsx
theodur: 💯

theodur (dismissed):

theodur (approved):

Sparowhawk (approved):

Issue #6586, opened 2023-08-24
CU - SPIKE - Tanstack/React-Query Worthiness


dumathane:
## Proposed Change

Currently the FE team is facing a large number of new API endpoints being introduced as part of the User Endpoint Improvements epic, should we take this time to revisit our process for state management and api loading mechanics and potentially replace them with a cleaner alternative.

## Why Should We Prioritize?

Now is a potentially appropriate time for lots of changes since a substantial amount are coming either way.

## Coding Time Estimation

Timeboxing the spike to a day in length with a report to the FE team about findings.	

## Testing Considerations

No applicable testing.

## Checklist
[ ] Add the front-end label tag

[ ] Attach to the Frontend epic for engineering initiatives

theodur:
# Overview
React Query is focused on managing server state (asynchronous state being fetched from the API). There is no need for global state management; this is instead replaced by queries that are cached. A common theme seems to be to use React Query for asynchronous data, while continuing to use traditional state management tools like Redux for synchronous state. One of the main goals of React Query seems to be reducing redundant patterns that are involved with using traditional state management tools.

# Queries
Queries are a way of mapping asynchronous data sources to a unique key, so they can be referenced in multiple places. The key is used to cache the asynchronous data, so that it's clear what the cache for that data is for. This is done via the `useQuery` hook, which accepts a name for the unique key, and the function that fetches the data. Here's an example from the [documentation](https://tanstack.com/query/v3/docs/react/guides/queries#query-basics):

```js
import { useQuery } from 'react-query'

function App() {
  const info = useQuery('todos', fetchTodoList)
}
```

The nice thing about `useQuery` is it handles a lot of things we're currently doing manually in our Redux store. States in our store relating to loading, like `loading`, `preloadComplete`, `needsDataLoad`, `error`, etc. are handled with React Query. From the `useQuery` hook, we get access to fetching information like `isLoading`, `isError`, etc. which allows us to reference the fetching state without having to manually declare it like we currently do in our store. This would be particularly useful for slices like [secure messaging](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/src/store/slices/secureMessagingSlice.ts), where there are multiple loading states in the slice (`loadingInbox`, `loadingFolders`, etc). From the example above from the docs, `info` would have attributes like `info.isLoading`, `info.isError`, etc that keep track of the state of fetching the data for us.

[...]

PR #6627, opened 2023-09-01
feature/6408-theo-DemographicsEndpointReactQuery

theodur:
## Description of Change

This is a redo of #6563 that uses React Query instead. For more information on React Query, you can read the [write up](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/6586#issuecomment-1698316753) I wrote. A lot of the file changes are just updating the profile object in test and mock data to ensure gender identity and preferred name isn't being pull from profile in demo mode.

## Screenshots/Video

N/A

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/api/queryClient.ts
```diff
@@ -0,0 +1,20 @@
...
+        await setAnalyticsUserProperty(analyticsUserProperty)
+      }
+    },
+    onError: (error, query) => {
+      const errorName = query?.meta?.errorName as string
+      if (errorName && isErrorObject(error)) {
+        logNonFatalErrorToFirebase(error, errorName)
+      }
+    },
+  }),
```
theodur, line 19: The query cache can access data from the `meta` option in the `useQuery` hook. This allows us to reduce redundancy in queries by adding common function calls.

The `useQuery` hook also has `onSuccess` and `onError` options where we can call common post API call functions like `logNonFatalErrorToFirebase` and `setAnalyticsUserProperty` from, but those options are going to be removed from the hook in the next major release (v5), so using the query cache is the recommended alternative.

Here's an example of how to pass data through the `meta` option:
https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6627/files#diff-e5f8dbfb8c5a6bd856e1ebd275a257052bc3503915bc6b4ead9530800a86d9cfR26

VAMobile/src/api/types/DemographicsData.ts
theodur, line 1: I moved these types here and removed them from `VAMobile/src/store/api/types/UserData.ts` since they're no longer being referenced or used in the Redux store. Future types used for queries or mutations can be added to the `VAMobile/src/api/types` directory also.

VAMobile/src/api/demographics/getDemographics.tsx
theodur, line 1: I made a new directory (`VAMobile/src/api`) for storing queries and mutations. Storing these in `VAMobile/src/store` didn't feel accurate, since these are in no way connected to the Redux store. I organized the queries and mutations in the `demographics` folder in separate files to make it easier to find. Future queries and mutations can be stored in their appropriate folder, e.g. `VAMobile/src/api/payments/getPayments`

VAMobile/src/components/Nametag/Nametag.test.tsx
```diff
@@ -20,12 +20,12 @@ context('Nametag', () => {
         personalInformation: {
           ...InitialState.personalInformation,
           profile: {
-            preferredName: 'Benny',
+            preferredName: '',
```
Sparowhawk, line 23: I know this might be nitpicking but is this and the genderIdentity field even needed anymore? Should we remove these as we go or will this be all removed once everything gets converted over. Lastly I think we forgot to ask how we mock the api call results for demo mode with the new way of doing things. Would it be the same?

VAMobile/src/screens/HomeScreen/ProfileScreen/PersonalInformationScreen/GenderIdentityScreen/GenderIdentityScreen.tsx
```diff
@@ -108,10 +100,19 @@ const GenderIdentityScreen: FC<GenderIdentityScreenProps> = ({ navigation }) =>
...
 
-  if (useError(ScreenIDTypesConstants.GENDER_IDENTITY_SCREEN_ID)) {
+  const onTryAgain = () => {
+    if (getDemographicsError) {
+      refetchDemographics()
+    }
+    if (getGenderIdentityOptionsError) {
+      refetchGenderIdentityOptions()
+    }
+  }
```
theodur, line 110: Since we're not relying on `useError` or setting the `tryAgain` function from the `errorSlice`, we need to manually set that here. This will probably be a common trend for screens that don't use the `useError` hook for determining whether an error happened, which I imagine will be a lot of our screens. FWIW, this screen is a unique case since there are 2 API calls that can error out. For screens dependent on only one API call, we would just pass the refetch function for that query directly into the `onTryAgain` prop for the error component

Sparowhawk (approved):

VAMobile/src/api/demographics/getGenderIdentityOptions.tsx
```diff
@@ -0,0 +1,40 @@
+import { useQuery } from '@tanstack/react-query'
+
+import { GenderIdentityOptions, GenderIdentityOptionsPayload } from '../types/DemographicsData'
+import { UserAnalytics } from 'constants/analytics'
+import { get } from '../../store/api'
```
narin, line 5: Relative imports

VAMobile/src/api/demographics/updateGenderIdentity.tsx
```diff
@@ -0,0 +1,53 @@
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+
+import { EditResponseData, put } from '../../store/api'
```
narin, line 3: Relative import

VAMobile/src/api/demographics/getGenderIdentityOptions.tsx
```diff
@@ -0,0 +1,40 @@
+import { useQuery } from '@tanstack/react-query'
```
narin, line 1: More of a general comment, but if we plan to have a separate file for every API call, we should think about naming conventions before we have 50+ query files. Otherwise we should consider grouping by feature.

narin (changes requested):
Looks good overall. Requested changes mainly around relative import nits. Have some questions about patterns moving forward.

VAMobile/src/api/demographics/getDemographics.tsx
```diff
@@ -0,0 +1,29 @@
...
+    throw error
+  }
+}
+
+/**
+ * Returns a query for user demographics
+ */
+export const useDemographics = () => {
+  return useQuery({
+    queryKey: ['user', 'demographics'],
```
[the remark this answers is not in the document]
theodur: Yeah, we definitely should. I manually declared the query keys because this started out as a POC, and I never got back to that. One of the maintainers of React Query recommends using a [Query Key Factory](https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories), which is essentially an object that contains the query keys for each _feature_. I think that might be a good way to go, since we could keep all the query keys in a separate file within the feature folder that can easily be referenced, e.g. `api/demographics/queryKeys.ts`.

Alternatively, we could store query keys as constants in the files of the queries using them, and just export the keys from there, but that might make it difficult to get an overview of all the queries for a feature. Plus, it'd be more difficult to do mass invalidations on multiple queries if they're in separate files.

The React Query [documentation](https://tanstack.com/query/v4/docs/react/community/lukemorales-query-key-factory) mentions a library for managing query keys, but I'm not sure that it's worth bringing in a separate library just for managing constants.

I'll probably go with storing the queries as constants in a separate file for now as recommended by one of the maintainers, unless anyone has other suggestions.

VAMobile/src/api/demographics/getDemographics.tsx
```diff
@@ -0,0 +1,29 @@
+import { useQuery } from '@tanstack/react-query'
+
+import { DemographicsPayload, UserDemographics } from '../types/DemographicsData'
```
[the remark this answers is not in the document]
theodur: Good eyes, VS Code auto generated the imports for me, so I didn't pay much attention to them. Not sure if we have a CU ticket for this, but it would be useful to have linting catch relative imports. When I was on VSP, we had a custom ESLint rule for that in the vets-website repo

VAMobile/src/api/demographics/getGenderIdentityOptions.tsx
theodur: Yeah I agree, I briefly touched on that in one of my [comments](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6627#discussion_r1316749733). What I was thinking is we could have a dedicated folder for each feature, and within those folders, have separate files named after the API call they're making, so it's easy to see where things are, which is how I have it structured in this PR. Alternatively, we could name the files after the query/mutation hooks that are exported, e.g. `useGenderIdentityOptions.tsx`. I'm curious how others feel

VAMobile/src/api/demographics/getGenderIdentityOptions.tsx
theodur: Fixed

VAMobile/src/api/demographics/updateGenderIdentity.tsx
theodur: Fixed

VAMobile/src/api/demographics/queryKeys.ts
```diff
@@ -0,0 +1,4 @@
+export const demographicsKeys = {
```
theodur, line 1: The query keys for demographics are slightly different from what a [typical query key factory](https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories) would look like. Typically, you would have an `all` property that holds the base key that other keys in the factory add to their array. This way, you could easily invalidate all queries related to a feature. For gender identity, I'm not sure that we have a case where we would need to invalidate both `demographics` and `genderIdentityOptions`, so I omitted `all` for this feature. If that changes, we can always add it here

VAMobile/src/api/demographics/getDemographics.tsx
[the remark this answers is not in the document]
dumathane: I like the factory concept.

VAMobile/src/api/demographics/getDemographics.tsx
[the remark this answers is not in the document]
dumathane: Searched import and relative and found nothing so [here's](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/6678) the new ticket.

VAMobile/src/api/demographics/getGenderIdentityOptions.tsx
dumathane: I think I like how Theo has it grouped currently, doesn't mean we couldn't change our minds on implementation though.

dumathane (approved):

Sparowhawk (approved):

VAMobile/src/api/demographics/getGenderIdentityOptions.tsx
Sparowhawk: I would agree with grouped by feature and then if a feature has a particular large amount of api calls then we can sub group them(secure messaging could be sub grouped to folders and messages for instance)

narin (approved):

PR #6635, opened 2023-09-01
bug/5068-Dylan-BSSVLettersLoadingFix

Sparowhawk:
## Description of Change
Updated the loading boolean for BSSV to be different from the letters list screen. This was what was causing both loadings to occur after each other.

## Screenshots/Video

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/b8e17f05-1b72-426f-84fe-8a83b61193da

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
It'd be nicer to only display the desired loading screen, and not the letters list one first.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/store/slices/lettersSlice.test.ts
+++ b/VAMobile/src/store/slices/lettersSlice.test.ts
@@ -207,9 +207,6 @@ context('letters', () => {
       await store.dispatch(getLetterBeneficiaryData())
       const actions = store.getActions()
 
-      const startAction = _.find(actions, { type: ActionTypes.LETTER_START_GET_BENEFICIARY_DATA })
-      expect(startAction).toBeTruthy()
-
       const endAction = _.find(actions, { type: ActionTypes.LETTER_FINISH_GET_BENEFICIARY_DATA })
       expect(endAction).toBeTruthy()
       expect(endAction?.state.letters.loading).toBe(false)
--- a/VAMobile/src/store/slices/lettersSlice.test.ts
+++ b/VAMobile/src/store/slices/lettersSlice.test.ts
@@ -207,6 +207,9 @@ context('letters', () => {
       await store.dispatch(getLetterBeneficiaryData())
       const actions = store.getActions()
 
+      const startAction = _.find(actions, { type: ActionTypes.LETTER_START_GET_BENEFICIARY_DATA })
+      expect(startAction).toBeTruthy()
+
       const endAction = _.find(actions, { type: ActionTypes.LETTER_FINISH_GET_BENEFICIARY_DATA })
       expect(endAction).toBeTruthy()
       expect(endAction?.state.letters.loading).toBe(false)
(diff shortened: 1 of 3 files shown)


VAMobile/src/store/slices/lettersSlice.test.ts
```diff
@@ -207,9 +207,6 @@ context('letters', () => {
       await store.dispatch(getLetterBeneficiaryData())
       const actions = store.getActions()
 
-      const startAction = _.find(actions, { type: ActionTypes.LETTER_START_GET_BENEFICIARY_DATA })
```
Sparowhawk, line 210: had to remove this part since the new timing of changing the loading variable to true after the api is called for the unit tests skipped this part. Wish we could mock a response time in order for this test to still work.

VAMobile/src/store/slices/lettersSlice.ts
```diff
@@ -79,10 +79,10 @@ export const getLetterBeneficiaryData =
   async (dispatch) => {
     dispatch(dispatchClearErrors(screenID))
     dispatch(dispatchSetTryAgainFunction(() => dispatch(getLetterBeneficiaryData(screenID))))
-    dispatch(dispatchStartGetLetterBeneficiaryData())
 
     try {
       const letterBeneficiaryData = await api.get<api.LetterBeneficiaryDataPayload>('/v0/letters/beneficiary')
+      dispatch(dispatchStartGetLetterBeneficiaryData())
```
theodur, line 85: After looking into this, it seems like the root of the issue is that both `dispatchStartGetLetterBeneficiaryData` and `dispatchStartGetLetters` (called from the previous screen) set the same `loading` state to `true`, which is causing the previous screen to load while navigating to the next screen. The reason moving the dispatch after the API call kind of works is because it gives the app a little more time to navigate away from the letters screen, so you don't see the loading component on that screen even though it's probably still being rendered.

Having a separate loading state for beneficiary data, maybe something like `loadingLetterBeneficiaryData` and updating `dispatchStartGetLetterBeneficiaryData` to set that instead, would fix the root issue. This is a great example of what React Query will help us avoid; needing to manually manage different loading state.

Also, while testing this, I think I found a new bug (unrelated to this ticket). Whenever the API call for fetching beneficiary data fails, the screen is stuck on the loading component. This is due to the [logic](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/123a0b450054423581e953e65874a962747f2958/VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx#L209-L215) for rendering the error component not accounting for when there's an error loading beneficiary data. I'm gonna spin up a new bug ticket for that

theodur (changes requested):

VAMobile/src/store/slices/lettersSlice.test.ts
Sparowhawk: removed

VAMobile/src/store/slices/lettersSlice.ts
Sparowhawk: updated to use this approach, can't wait to move this on over to react query lol

dumathane (approved):

PR #6643, opened 2023-09-05
CI e2e test automation and Android

timwright12:
Detox tests for Android in CI and QA automation to post to slack thread during daily QA builds

**From Dylan:**
Tickets created for iOS and Android fixing tests and turning slack feature on

https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/6471323790

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/8aab205d-0c04-4430-964b-5f361713faed

### DIFF
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -98,6 +98,14 @@ jobs:
       - name: Bundle Android app
         run: yarn e2e:android-build
 
+      - id: Get_AVD_Name
+        run: |
+          content=`cat ./path/to/package.json`
+          content="${content//'%'/'%25'}"
+          content="${content//$'\n'/'%0A'}"
+          content="${content//$'\r'/'%0D'}"
+          echo "::set-output name=packageJson::$content"
+
       - name: Run e2e tests for Android
         uses: reactivecircus/android-emulator-runner@v2
         with:
@@ -107,7 +115,7 @@ jobs:
           force-avd-creation: false
           disable-animations: true
           arch: x86_64
-          avd-name: Pixel_4_XL_API_28
+          avd-name: "${{fromJson(steps.Get_AVD_Name.outputs.packageJson).devices.emulator.device.avdName}}"
           script: yarn e2e:android-test --take-screenshots failing --updateSnapshot --headless
 
       # - name: Inform Slack of Failure
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -103,9 +103,9 @@ jobs:
         with:
           working-directory: VAMobile
           api-level: 28
+          force-avd-creation: false
+          disable-animations: true
           cores: 2
-          ram-size: 512M
-          heap-size: 2048M
           arch: x86_64
           avd-name: Pixel_4_XL_API_31
           script: yarn e2e:android-test --take-screenshots failing --updateSnapshot --headless
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -103,13 +103,12 @@ jobs:
         with:
           working-directory: VAMobile
           api-level: 28
-          profile: pixel
+          profile: Pixel 6 Pro
           force-avd-creation: false
           disable-animations: true
           arch: x86_64
           avd-name: Pixel_4_XL_API_28
-          script: avdmanager list device
-          # yarn e2e:android-test --take-screenshots failing --updateSnapshot --headless
+          script: yarn e2e:android-test --take-screenshots failing --updateSnapshot --headless
 
       # - name: Inform Slack of Failure
       #   id: post_failure
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -95,10 +95,6 @@ jobs:
       - name: Set app environment variables
         run: yarn env:staging
 
-      - name: Setup metro
-        working-directory: VAMobile
-        run: yarn start
-
       - name: Bundle Android app
         run: yarn e2e:android-build
 
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -98,9 +98,17 @@ jobs:
       - name: Bundle Android app
         run: yarn e2e:android-build
 
+      - name: Setup_JSON_File
+        run: |
+          content=`cat ./path/to/package.json`
+          content="${content//'%'/'%25'}"
+          content="${content//$'\n'/'%0A'}"
+          content="${content//$'\r'/'%0D'}"
+          echo "::set-output name=packageJson::$content"
+
       - name: Get_AVD_Name
         working-directory: VAMobile
-        run: echo AVD_Name="${{fromJson(cat ./.detoxrc.json).devices.emulator.device.avdName}}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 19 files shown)


timwright12:
@Sparowhawk can you get a passing CI run before calling this done? We have everything passing locally, so I don't think we can just say all the tests are broken

.github/workflows/e2e_android.yml
```diff
@@ -0,0 +1,138 @@
...
+# Detox e2e tests in CI
+#
+
+name: '[e2e] Android Detox Test Run'
+
+on:
+  workflow_dispatch:
+  # Run on weekdays 4:00 AM UTC
+  schedule:
+    - cron: '0 4 * * 1,2,3,4,5'
```
timwright12, line 11: Let's take this out until it actually passes. No need to ping slack or run cron just for it to fail

Sparowhawk: @timwright12 I have taken it out. However these tests are not passing locally. I confirmed with @alexandec and @rbontrager that a decent number for iOS and Android are broken. Also unfortunately once one test breaks it breaks all the tests that occur after it.

I have moved this back to draft and ticket https://github.com/department-of-veterans-affairs/va-mobile-app/issues/6879 is created for fixing the broken e2e tests that I found

timwright12: Might want to hop on a zoom with @rbontrager to investigate your set up. Every test outside of appointments was passing locally at last check

.github/workflows/e2e_android.yml
```diff
@@ -106,7 +106,7 @@ jobs:
           force-avd-creation: false
           disable-animations: true
           arch: x86_64
-          avd-name: Pixel_4_XL_API_31
+          avd-name: Pixel_4_XL_API_28
```
timwright12, line 109: This is in a local JSON file so we should just be able to pull it in so there's no need to keep it in sync https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/.detoxrc.json#L28

.github/workflows/e2e_android.yml
Sparowhawk: @timwright12 We were able to update detox and the platform error is no longer occurring. The issue now is that locally we have a different resolution/screen size than what we can enable on the emulator/simulator. So @rbontrager is now adding some scrolling to get most of the remaining tests to pass. We also came across a problem when changing json mock data in the files as well that seems to be a detox issue.

.github/workflows/e2e_android.yml
Sparowhawk: added

Sparowhawk: after playing with this for a while I was able to extra the name from the json file, but then this block of code no longer runs with this error:

Creating AVD.
  /bin/sh -c \echo no | avdmanager create avd --force -n Pixel_4_XL_API_28 --abi 'default/x86_64' --package 'system-images;android-28;default;x86_64' --device 'pixel_4_xl'
  Error: Package path (-k) not specified. Valid system image paths are:
  system-images;android-28;default;x86_64
  null
  
  So reverting back to the hard coded way that works as I tried trimming white space and everything else that might be going on. The command from what I can tell in the logs is the exact same and hard coded works and fetching from the file does not

Sparowhawk:
@timwright12 This is now ready for review, I was able to pull the name from the json file, however by doing so the commands for that code block failed even though they were the exact same as the hard coded value, so I resorted to going back to hard coding that value. 

I have created/updated the tickets for the 6 tests for iOS and the 8 tests for Android along with the 2 tickets to turn on the slack feature for Android and iOS respectively once they are working as this ticket is now going greatly over scope.

.github/workflows/e2e_android.yml
```diff
@@ -0,0 +1,140 @@
...
+#
+
+name: '[e2e] Android Detox Test Run'
+
+on:
+  workflow_dispatch:
+  # Run on weekdays 4:00 AM UTC
+  # schedule:
+  #   - cron: '0 4 * * 1,2,3,4,5'
+  push:
```
timwright12, line 12: Remove before merging

.github/workflows/e2e_android.yml
```diff
@@ -0,0 +1,140 @@
...
+          bundler-cache: true
+          working-directory: VAMobile/android
+
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: 17
+          distribution: 'microsoft'
+      
+      - name: Setup Android SDK
```
timwright12, line 82: Is this duplicative with reactivecircus/android-emulator-runner@v2?

.github/workflows/e2e_ios.yml
```diff
@@ -6,6 +6,10 @@ name: '[e2e] iOS Detox Test Run'
 
 on:
   workflow_dispatch:
+  # Run on weekdays 4:00 AM UTC
+  # schedule:
+  #   - cron: '0 4 * * 1,2,3,4,5'
+  push:
```
timwright12, line 12: Remove before merging

VAMobile/.detoxrc.json
```diff
@@ -19,13 +27,13 @@
     "simulator": {
       "type": "ios.simulator",
       "device": {
-        "type": "iPhone 13"
+        "type": "iPhone 13 Pro Max"
```
timwright12, line 30: Confirm with QA that this change is OK
@rbontrager

.github/workflows/e2e_ios.yml
```diff
@@ -16,9 +20,20 @@ env:
   APP_CLIENT_SECRET: ${{ secrets.APP_CLIENT_SECRET }}
   # IAM production app client secret
   APP_CLIENT_SECRET_PROD: ${{ secrets.APP_CLIENT_SECRET_PROD }}
+  # Slack API token used by fastlane
+  SLACK_API_TOKEN: ${{ secrets.SLACK_API_TOKEN }}
```
timwright12, line 24: @Sparowhawk this stuff should be in a separate ticket/PR, we're not alerting Slack yet until all the runs pass. Can you break it out? 

Also shouldn't be any iOS changes in this PR since it's an android ticket

.github/workflows/e2e_ios.yml
Sparowhawk: removed

.github/workflows/e2e_android.yml
Sparowhawk: removed

.github/workflows/e2e_ios.yml
Sparowhawk: removed the iOS changes

timwright12:
Did we get a passing run in CI? Doesn't have to be all the tests, but at least 1 suite passing so we know we did it correctly

Sparowhawk:
> Did we get a passing run in CI? Doesn't have to be all the tests, but at least 1 suite passing so we know we did it correctly

@timwright12 Yes we did, several individual tests passed. The ones that didn't have tickets for and are likely fixed by Rachel's iOS fixes in her other branch

timwright12:
Can you link up to the workflow that passed in the description so we have that artifact?

Sparowhawk:
@timwright12 added link and video of the passing and failing e2e tests with the workflow

timwright12:
thanks, sounds good. I cant approve this because it was my PR originally. Whenever @rbontrager is comfortable, I'm good

Sparowhawk (approved):

PR #6711, opened 2023-09-12
added dev branch trigger to documentation workflow

timwright12:
The current documentation deploy happens daily and when we click a button, this is just adding an automatic deploy upon merge into `develop`


.github/workflows/documentation_deploy.yml
```diff
@@ -3,6 +3,8 @@ name: '[Documentation] Deploy Site'
 # Adapted from https://docusaurus.io/docs/deployment
 
 on:
+  push:
+    branches: ["develop"]
```
theodur, line 7: Would it be useful to use a [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) here so that this only runs when there are changes in `VAMobile/documentation`?

timwright12: yeah good call out, lemme update that

timwright12: added, hard to test, but I think I got the syntax

alexandec (approved):

narin (approved):

PR #6733, opened 2023-09-14
feature/6410-theo-contact-info-React-Query

theodur:
## Description of Change

Uses the new contact info endpoint for retrieving user contact information in React Query.

## Screenshots/Video

N/A

## Testing

Tested locally and on staging.

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/api/contactInformation/validateAddress.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const useValidateAddress = () => {
+  return useMutation({
+    mutationFn: validateAddress,
+    onError: async (error) => {
+      if (isErrorObject(error)) {
+        logNonFatalErrorToFirebase(error, 'validateAddress: Service error')
+      }
+    },
+  })
+}
```
theodur, line 35: This could easily be a query, and conceptually it would probably make more sense since this isn't mutating anything. The issue I ran into was handling when to have the query called and passing the data into the query. Because of how the `EditAddressScreen` is set up, even if the query is disabled to run on mount (we only want it to run when a user taps `Save`), the query needs to be manually called with the `refetch` function, but we can't pass parameters for the `queryFn` to the`refetch` function, so we'd need to pass in a dynamic state variable for handling changes to the address from `getAddressValues`. Although it's possible, it turned out to be a little more cumbersome than just having this be a mutation and prop drilling this to the `AddressValidation` component. Might have to think on refactoring this to a query as it'll avoid the need for prop drilling, and we won't have to pass the `mutate` function into the `AddressValidation` component

VAMobile/src/api/demographics/updateGenderIdentity.tsx
```diff
@@ -42,9 +39,9 @@ export const useUpdateGenderIdentity = () => {
       queryClient.invalidateQueries({ queryKey: demographicsKeys.demographics })
       showSnackBar(snackbarMessages.successMsg, dispatch, undefined, true, false, true)
     },
-    onError: async (data: string, error) => {
+    onError: async (error, variables) => {
```
theodur, line 42: Unrelated to this PR, but I noticed I switched up the params in my initial PR

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/AddressValidation/AddressValidation.tsx
```diff
@@ -22,17 +23,22 @@ export type AddressValidationProps = {
   addressEntered: AddressData
   addressId: number
   snackbarMessages: SnackbarMessages
+  validationData: ValidateAddressData
+  updateAddress: UseMutateFunction<EditResponseData | undefined, unknown, AddressData, unknown>
```
theodur, line 27: Unfortunately, I had to pass the same mutate function from the `EditAddressScreen` so that it can know when to display the loading component when the address is being validated or updated. Even though we could make a new instance of the `useValidateAddress` hook in this component, it doesn't seem like we have a good way to get the loading state of that mutation from the `EditAddressScree`. There is the [useIsMutating](https://tanstack.com/query/latest/docs/react/reference/useIsMutating) hook, but that only returns the number of matching mutation, but no information on the mutations themselves. This could be changing in v5 thought I believe

VAMobile/src/utils/personalInformation.tsx
```diff
@@ -97,14 +93,3 @@ export const getAddressDataFromSuggestedAddress = (suggestedAddress: SuggestedAd
...
-export const getPhoneDataForPhoneType = (phoneType: PhoneType, profile: UserDataProfile): PhoneData | undefined => {
-  switch (phoneType) {
-    case PhoneTypeConstants.HOME:
-      return profile.homePhoneNumber
-    case PhoneTypeConstants.MOBILE:
-      return profile.mobilePhoneNumber
-    case PhoneTypeConstants.WORK:
-      return profile.workPhoneNumber
-  }
-}
```
theodur, line 110: This function is just checking if the user has a `phoneType` saved, and returns it. This was only being used in the `deleteEmail` Redux action; right before making the API call it would call it to verify that the user had the phone type save in their profile. `EditPhoneNumberScreen` gets passed the `phoneData` prop which comes from the `contactInformation`. So just checking if `phoneData` isn't null should mean that the user has it saved on their profile. Not sure that this function call is needed to verify that

VAMobile/src/api/contactInformation/updateAddress.tsx
```diff
@@ -0,0 +1,42 @@
...
+    onSuccess: async () => {
+      await setAnalyticsUserProperty(UserAnalytics.vama_uses_profile())
+      queryClient.invalidateQueries({ queryKey: contactInformationKeys.contactInformation })
+    },
+    onError: async (error) => {
+      if (isErrorObject(error)) {
+        logNonFatalErrorToFirebase(error, 'updateAddress: Service error')
+      }
+    },
+  })
```
theodur, line 41: I made a single mutation for both creating and updating an address, since the code is basically identical. In contrast, I made separate mutations for [creating](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6733/files#diff-bfe1e2437d52a629b2fdbc0e1845ee616c779ba5416f05f02bb98593b6319f9eR1) and [updating](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6733/files#diff-e46a82603c8f50a860943a1931db8cf6262def93a6f0b80acef6e78f5bb0978bR1) phone numbers. I wanted to show what both approaches look like and see if there's a strong preference for which pattern we should adapt. The pros about using one mutation for both creating and updating is fewer lines of code. On the other hand, we lose the ability to differentiate between when the entity is being updated or created, since there would only be one `isLoading` status for the single mutation. Also, semantically, the naming of a single mutation hook might not be as clear to what it does. You can see both approaches in action in the [EditAddressScreen](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6733/files#diff-5289db0ef9e10d828712591d88415c37849600f4bed6b3b0a6d48ee0970b182aR281) and [EditPhoneNumberScreen](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6733/files#diff-8d68d775f5241ccbc8a26649b596afdf7732e575d7129ddd5c42cf33bbb97402R109-R130) files

Sparowhawk: I personally like having them be in the same mutation as opposed to separate mutation files

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/EditPhoneNumberScreen/EditPhoneNumberScreen.tsx
```diff
@@ -90,9 +88,46 @@ const EditPhoneNumberScreen: FC<IEditPhoneNumberScreen> = ({ navigation, route }
...
+        id,
+      }
+
+      const update = () =>
+        updatePhoneNumber(phoneDataPayload, {
+          onSuccess: () => showSnackBar(saveSnackbarMessages.successMsg, dispatch, undefined, true, false, true),
+          onError: () => showSnackBar(saveSnackbarMessages.errorMsg, dispatch, update, false, true, true),
+        })
+
+      update()
```
alexandec, line 121: If this is the only place the function is used, probably don't need a function

alexandec (dismissed):
Added some optional suggestions/comments, looks good overall

Sparowhawk (approved):

Sparowhawk (dismissed):

alexandec (dismissed):

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/EditPhoneNumberScreen/EditPhoneNumberScreen.tsx
theodur: It's also used in the `showSnackBar` function, because it gets used as the retry function on errors. That's the main reason I wrapped that logic in a separate function, to avoid needing to pass in the parameters to the mutation function twice. I have [variations](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6733/files#diff-4b0dfaf60714e697abefaf28a855abc1523660a9c20139bce60f4d65a3ba2e54R124-R128) of this where I store the mutate options in a separate variable, and just pass that into the functions. Mainly experimenting with different styles to see which one feels the most natural

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/EditPhoneNumberScreen/EditPhoneNumberScreen.test.tsx
```diff
@@ -31,180 +27,75 @@ context('EditPhoneNumberScreen', () => {
...
     describe('when the length is less than or equal to 10 digits', () => {
-      it('should display just the numbers in the text input', async () => {
-        const phoneNumTextInput = testInstance.findAllByType(TextInput)[0]
-        phoneNumTextInput.props.onChangeText('12345')
-        expect(phoneNumTextInput.props.value).toEqual('12345')
+      it('displays just the numbers in the text input', async () => {
+        const phoneNumberInput = screen.getByTestId('phoneNumberTestID')
+
+        fireEvent.changeText(phoneNumberInput, '12345')
+        phoneNumberInput.props.onEndEditing()
```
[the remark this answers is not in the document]
theodur: I feel the same way. When I was working on this, I initially tried to interact with just the UI, but the formatting changes to the number wouldn't happen unless the `onEndEditing` prop was fired off 🙁. I tried tapping into other elements after changing the phone number input, but still no luck. In `EditPhoneNumberScreen`, the formatting happens once `onEndEditing` is called, so it seems that might not be happening with `fireEvent` for some reason. I wonder if it would work with [user-event](https://github.com/testing-library/user-event) library, which is what RTL recommends for simulating user events

VAMobile/src/testUtils.tsx
```diff
@@ -233,8 +233,12 @@ ctxFn.skip = (name: string, fn: () => void) => {
 
 export const context = ctxFn
 
+export type QueriesData = Array<{
+  queryKey: QueryKey
+  data: any
+}>
 //@ts-ignore
-function render(ui, { preloadedState, navigationProvided = false, ...renderOptions } = {}) {
+function render(ui, { preloadedState, navigationProvided = false , ...renderOptions } = {}, queriesData?: QueriesData) {
```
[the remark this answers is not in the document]
theodur: Yeah I agree, originally I had it within the object in the 2nd param, but something threw me off with the types, so I moved it to the end of the function. It definitely makes it easier to read and cleaner having it there though. I just moved it back 👍. And initially I had it as a single object named `queryData` instead of an array of objects, but decided to make an array, so it can handle setting up data for multiple queries just in case. Now that I think about it, we could also update it to be a single object where the `key` is the query key and the `value` is the data

VAMobile/src/testUtils.tsx
```diff
@@ -231,8 +231,19 @@ ctxFn.skip = (name: string, fn: () => void) => {
 
 export const context = ctxFn
 
+export type QueriesData = Array<{
+  queryKey: QueryKey
+  data: any
+}>
+
+type RenderParams = {
+  preloadedState?: any // TODO: Update this type to Partial<RootState> and fix broken tests
```
theodur, line 240: Some of the tests are broken when the type is set for this because of incorrectly formatted state. This parameter may not even be needed once we complete the React Query migration, but I left a comment here just in case

VAMobile/src/api/contactInformation/getContactInformation.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Returns a query for user contact information
+ */
+export const useContactInformation = () => {
+  return useQuery({
+    queryKey: contactInformationKeys.contactInformation,
+    queryFn: () => getContactInformation(),
+    staleTime: 5000,
```
theodur, line 32: I adjusted the staleTime to 5 seconds because the `ContactInformationScreen` renders the `AddressSummary` component, and both components use this query, which results in 2 calls being made to the contact info endpoint. 5 seconds is enough time to prevent the query from refetching in the child component, and also gives the user some time to select what info they want to edit, since the editing screens also use the query, and would refetch the data.

I manually set the stale time in this query for now, but this touches on an important issue on how we want to manage stale time for our queries, because we can also set a global stale time in the query client, and update the stale time for individual queries when needed. React Query's default stale time of 0 may be a bit aggressive for our app

Sparowhawk (dismissed):

DJUltraTom (dismissed):

Sparowhawk (approved):

alexandec (approved):

PR #6743, opened 2023-09-14
bug/5401-chris-add-snackbar-null-events

alexandec:
## Description of Change

Log a new event, with location, when snackBar is null and snackBar.hideAll() is called. This will tell us:
- whether snackBar.hideAll() is still being called while snackBar is null
- where in the app these calls are occurring, and how often

Our ultimate goal is to track down the exact cause of the issue and fix it.

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

Event is sent when snackBar is null and snackBar.hideAll() is called

## PR Checklist

## For QA


Sparowhawk (approved):

narin (approved):

alexandec:
Checked with Tom and there's nothing to QA on this one. Merging it

PR #6769, opened 2023-09-19
feature/6651-Dylan-userLoggedInApiPost

Sparowhawk:
## Description of Change
Added post to /v0/user/logged-in on successful login. So there is no handling needed for success or failure.

## Screenshots/Video
N/A

## Testing
Yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Acceptance Criteria:
Mobile app makes POST request to v0/user/logged-in after successful login.

## PR Checklist

## For QA


VAMobile/src/api/loggedIn/postLoggedIn.tsx
```diff
@@ -0,0 +1,8 @@
+import { post } from 'store/api'
+
+/**
+ * Post user logged in
+ */
+export const postLoggedIn = async () => {
+  await post('v0/user/logged-in')
```
alexandec, line 7: Shouldn't this be `/v0/user/logged-in` with a leading slash?

Sparowhawk: good catch on the typo. fixed

VAMobile/src/store/slices/authSlice.ts
```diff
@@ -807,6 +808,9 @@ const authSlice = createSlice({
     dispatchFinishAuthLogin: (state, action: PayloadAction<AuthFinishLoginPayload>) => {
       const successfulLogin = !action.payload.error
 
+      if (successfulLogin) {
+        postLoggedIn()
+      }
```
theodur, line 813: Since this is a reducer, we shouldn't be performing any async actions here because this should be a pure function. I'd suggest calling this in the thunks, right before the `dispatchFinishAuthLogin` calls are made

theodur (changes requested):

VAMobile/src/store/slices/authSlice.ts
Sparowhawk: updated also added logging errors to the api

narin (approved):

VAMobile/src/store/slices/authSlice.ts
```diff
@@ -699,6 +700,7 @@ export const handleTokenCallbackUrl =
       await logAnalyticsEvent(Events.vama_login_success(SISEnabled))
       await dispatch(dispatchSetAnalyticsLogin())
       dispatch(dispatchFinishAuthLogin({ authCredentials }))
+      usePostLoggedIn()
```
theodur, line 703: Since this isn't in a component, we shouldn't be using a hook here, we can directly call the `postLoggedIn` function. Also, since `postLoggedIn` is only used here, we probably don't need to store it in a separate file. It doesn't seem like something we'll need for other parts of the app

theodur (changes requested):

VAMobile/src/store/slices/authSlice.ts
Sparowhawk: updated

narin (approved):

theodur (dismissed):

theodur (approved):

PR #6782, opened 2023-09-20
Fix focus on error issue in Messaging

alexandec:
## Description of Change

This PR fixes the focus on error issue we had in Messaging. The cause was that the FullScreenSubtask template was being rendered without passing the scrollViewRef while the screen was loading, then rendered again with the scrollViewRef passed. For some reason (possibly internal to how React works), if the ref isn't passed on the first render, it doesn't get passed on the second even though we attempt to pass it. The fix is to always pass the ref.

## Screenshots/Video

Start new message (before)

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/5805150/8f5d5eb9-6aaf-401b-969a-fcee5b2d50ba

Start new message (after)

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/5805150/99b6e155-d04a-4a31-8e01-d86ef52add64

Reply (before)

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/5805150/b6e44eee-041c-411c-93dc-94c29314e4b6

Reply (after)

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/5805150/4b265319-2072-46a5-b11a-1f0c65025da9

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

On error, focus should shift to the error alert box, which would announce to the screen reader that there IS a problem & how to fix it.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HealthScreen/SecureMessaging/EditDraft/EditDraft.tsx
+++ b/VAMobile/src/screens/HealthScreen/SecureMessaging/EditDraft/EditDraft.tsx
@@ -288,7 +288,7 @@ const EditDraft: FC<EditDraftProps> = ({ navigation, route }) => {
 
   if (useError(ScreenIDTypesConstants.SECURE_MESSAGING_COMPOSE_MESSAGE_SCREEN_ID)) {
     return (
-      <FullScreenSubtask title={t('editDraft')} leftButtonText={t('cancel')} menuViewActions={MenViewActions}>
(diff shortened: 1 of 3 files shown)


theodur (approved):
So strange, this was definitely working before without passing the `scrollViewRef` to the other instances of the template. Not sure what changed 🤔 It's probably safer to pass the prop to the other instances anyway though

PR #6872, opened 2023-09-28
finished

Sparowhawk:
## Description of Change
Changed the screen reader analytics from firing on app launch to firing on app launch and when screen reader is turned on/off

## Screenshots/Video
N/A

## Testing
Yarn Test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
App analytics tracks all screen reader users regardless if they turn it on before or after opening the app.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/App.tsx
+++ b/VAMobile/src/App.tsx
@@ -40,7 +40,7 @@ import { isIOS } from 'utils/platform'
 import { linking } from 'constants/linking'
 import { profileAddressType } from 'screens/HomeScreen/ProfileScreen/ContactInformationScreen/AddressSummary'
 import { updateFontScale, updateIsVoiceOverTalkBackRunning } from './utils/accessibility'
-import { useAppDispatch } from 'utils/hooks'
+import { useAppDispatch, useIsScreenReaderEnabled } from 'utils/hooks'
 import { useColorScheme } from 'styles/themes/colorScheme'
 import { useHeaderStyles, useTopPaddingAsHeaderStyles } from 'utils/hooks/headerStyles'
 import BiometricsPreferenceScreen from 'screens/BiometricsPreferenceScreen'
@@ -164,6 +164,7 @@ export const AuthGuard: FC = () => {
   // This is to simulate SafeArea top padding through the header for technically header-less screens (no title, no back buttons)
   const topPaddingAsHeaderStyles = useTopPaddingAsHeaderStyles()
   const [currNewState, setCurrNewState] = useState('active')
+  const screenReaderEnabled = useIsScreenReaderEnabled()
 
   const snackBarProps: Partial<ToastProps> = {
     duration: SnackBarConstants.duration,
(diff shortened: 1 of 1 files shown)


VAMobile/src/App.tsx
```diff
@@ -188,10 +189,9 @@ export const AuthGuard: FC = () => {
   }, [isVoiceOverTalkBackRunning, dispatch, currNewState])
 
   useEffect(() => {
-    // only run on app load
     dispatch(sendUsesLargeTextAnalytics())
     dispatch(sendUsesScreenReaderAnalytics())
-  }, [dispatch])
+  }, [dispatch, screenReaderEnabled])
```
narin, line 194: Shouldn't `screenReaderEnabled` only be added to this array if it's used within the `useEffect()`?

theodur: It looks like `screenReaderEnabled` is needed in the array in order for analytics to be sent each time the screen reader is turned off/on, but Dylan can probably confirm that. Although, with this update, we'd be sending large text analytics every time the screen reader is turned off/on which isn't ideal.  We may need to move `sendUsesLargeTextAnalytics` into its own effect to prevent duplicate large text analytics from being sent

Sparowhawk: @narin @theodur That is correct it needs to be sent each time the screen reader is turned on/off so you have to add it to the array to trigger as dispatch doesn't change.

SendUsesLargeTextAnalytics has the same issue as screen reader in that it is only sent on app launch. So it needs to be included as the same issues apply and I don't see a good hook that we can use to send that separately.

dumathane: I think Theo is implying you make a second useEffect here and get a usesLargeText trigger for that second clause so we aren't sending usesLargeTextAnalytics anytime they turn off/on the screenreader? Is that accurate Theo?

theodur: @dumathane Yup, that's spot on. Large text analytics probably shouldn't be sent whenever a user enables/disables the screen reader, because it's unrelated. Having it in its own hook would ensure it's only sent when necessary, even if that is only when the app is first launched. Maybe the [useFontScale](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/b06a68ccfe27d1f9538fc879ce1e1964aafe221e/VAMobile/src/utils/hooks/index.tsx#L66-L80) would be a good hook to use for the new useEffect's dependency array, since it'd tell us when the font scale changes. 

Although, it looks like large text analytics are out of the scope of this ticket? It's still might be worth using `useFontScale` since large text analytics will need its own effect anyway

VAMobile/src/App.tsx
```diff
@@ -164,6 +164,9 @@ export const AuthGuard: FC = () => {
   // This is to simulate SafeArea top padding through the header for technically header-less screens (no title, no back buttons)
   const topPaddingAsHeaderStyles = useTopPaddingAsHeaderStyles()
   const [currNewState, setCurrNewState] = useState('active')
+  const screenReaderEnabled = useIsScreenReaderEnabled()
+  const fontScaleFunction = useFontScale()
+  const sendUsesLargeTextScal = fontScaleFunction(30)
```
theodur, line 169: Looks like there might be a typo in the variable name? Not sure what `Scal` stands for at the end.

It'd be ideal if we had a dedicated hook for determining when large text is using, similar to how we have `useIsScreenReaderEnabled`, but that's way beyond the scope of this ticket, and this should work for now

theodur (approved):

PR #6982, opened 2023-10-11
bug/6980-theo-GenderIdentityNotLoading

theodur:
## Description of Change

When the Personal Information screen is loaded, the user's gender identity isn't being shown. This is due to the gender identity options not being fetched, which results in the screen not knowing what option to map the user's gender identity to. This is happening because the `initialData` option was accidentally set for the `useGenderIdentityOptions` query, so when the `staleTime` was enabled for all queries in https://github.com/department-of-veterans-affairs/va-mobile-app/pull/6733, it doesn't mark the initial data as stale until 5 seconds after the query with the initial data was fetched. Removing the `initialData` resolves this issue.

Other changes in this PR include some minor cleanup/updates, and new unit tests to verify gender identity is being shown appropriately.

### Before

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/9042882/ed984f57-c2bd-4f93-9997-3813efdbfc43

### After

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/9042882/b74da4b5-add5-4997-bc68-14045aa52d1d

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

- [ ] Gender Identity data is shown when the personal information screen it loaded

## PR Checklist

## For QA


VAMobile/src/screens/HomeScreen/ProfileScreen/PersonalInformationScreen/PersonalInformationScreen.tsx
```diff
@@ -89,10 +89,10 @@ const PersonalInformationScreen: FC<PersonalInformationScreenProps> = ({ navigat
     }
   }
 
-  if (screenError || getDemographicsError || getGenderIdentityOptionsError || getPersonalInfoError) {
+  if (personalInformationInDowntime || getDemographicsError || getGenderIdentityOptionsError) {
```
theodur, line 92: I removed `getPersonalInfoError` as a condition for showing the error component because this screen is mainly for giving user's access to edit demographics data. The only way information from the `v2/user` endpoint is used is in displaying the DOB, and when that information isn't available, a message under the DOB tab is displayed saying this information is unavailable. 

Plus, chances are by the time a user navigates to this screen, the user's personal information has already been fetched from the home screen, so even if the personal info query were to fail here, we'd still have the previously fetched personal info stored in cache, and that would be used for the DOB

Sparowhawk (approved):

dumathane (approved):

PR #7181, opened 2023-10-31
bug/6981-chris-duckduckgo-login-failure

alexandec:
## Description of Change

Currently we always use the default browser when launching the Custom Tab for login. Since some browsers don't support Custom Tabs, this PR checks the installed browsers and chooses one that supports Custom Tabs. It works as follows:

1. If the user's default browser supports Custom Tabs, use that browser
2. If the user's default browser does not support Custom Tabs, find a different installed browser that does support Custom Tabs, and use that browser instead
3. If none of the installed browsers support Custom Tabs, use the default browser (best we can do)

Some browsers that support Custom Tabs: Chrome, Firefox, Brave, Edge, Samsung Internet Browser
Some browsers that don't support Custom Tabs: DuckDuckGo, Opera Mini

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

Test with a variety of default browsers, and a variety of browsers installed and not installed. Behavior should match what is described above.

Chrome can't be uninstalled but it can be disabled, which simulates being uninstalled.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
+++ b/VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
@@ -1,10 +1,12 @@
 package gov.va.mobileapp.native_modules
 
 import android.content.Intent
-import android.content.Intent.*
+import android.content.Intent.ACTION_VIEW
+import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.content.pm.PackageManager
 import android.net.Uri
 import androidx.browser.customtabs.CustomTabColorSchemeParams
+import androidx.browser.customtabs.CustomTabsClient
 import androidx.browser.customtabs.CustomTabsIntent
 import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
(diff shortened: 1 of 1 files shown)


Sparowhawk (approved):

VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
```diff
@@ -84,13 +86,27 @@ class CustomTabsIntentModule(private val context: ReactApplicationContext) :
...
+            // https://developer.chrome.com/docs/android/custom-tabs/howto-custom-tab-check/
+            val intent = Intent(ACTION_VIEW, Uri.parse("https://www.example.com"))
+            val intentHandlers = context.packageManager.queryIntentActivities(intent, PackageManager.MATCH_ALL)
+            val intentHandlerPackages = intentHandlers.map {
+                it.activityInfo.packageName
+            }.toTypedArray().asList()
+
+            // From the above list, find a package that supports Custom Tabs. If no installed
+            // packages support Custom Tabs, packageName will be set to null here
+            val packageName = CustomTabsClient.getPackageName(context, intentHandlerPackages)
```
TimRoe, line 99: Not clear by the comment: this will preferentially choose their default browser as described by the PR write-up, right? Wanted to double check it wouldn't arbitrarily just pull the first in the list and end up doing it alphabetically or whatever the prior list comes as.

narin (approved):

VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
```diff
@@ -84,13 +86,27 @@ class CustomTabsIntentModule(private val context: ReactApplicationContext) :
...
+
+            // If packageName is present, that package supports Custom Tabs, so use it
+            // TODO: Inform user when no installed packages support Custom Tabs
```
TimRoe, line 102: Did this already have a ticket number to point to? If so, may be good to list.

VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
TimRoe: Answered my own question following the Chrome Dev link above--yes, getPackageName does preferentially choose default unless told not to.

Defer to you if it's worth clarifying the comment.

TimRoe (approved):

theodur (approved):

VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
alexandec: There's no ticket on this one yet. I think there are some open questions on exactly how and when to notify users, in addition to the technical considerations of whether this happens in native code or not. Probably we'd need some input beyond engineers. There's also the wider consideration of whether we continue to refine the custom tabs approach or whether we move towards in-app login

VAMobile/android/app/src/main/java/gov/va/mobileapp/native_modules/CustomTabsIntentModule.kt
alexandec: Exactly, the third argument to getPackageName determines that. I tested a bit with multiple browsers installed. My choice of default browser was respected as long as it supported custom tabs. So it seems to work as advertised. QA will test further to make sure

DJUltraTom (approved):

PR #7192, opened 2023-11-02
feature/7189-rachael-github-actions-testRail-integration

rbontrager:
## Description of Change

Make a GitHub Actions script that sends the results of the automated tests to testRail

## Screenshots/Video

FullAutomatedTestTestRail

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


timwright12 (dismissed):

.github/workflows/update_testrail_results.yml
```diff
@@ -0,0 +1,96 @@
+name: '[e2e] Update TestRail Results'
+
+on:
+  push:
+    branches:
+      - 'dependabot/**'
```
theodur, line 6: The E2E workflows are already triggered for pushes to Dependabot branches, so this might be redundant. Instead of triggering this on Dependabot branches, we could use the `workflow_run` trigger to run this when the E2E iOS/Android workflows complete. Alternatively, we could also make this a reusable workflow and call this in the E2E workflows instead of making the E2E workflows reusable. The second option might be better because we wouldn't need to have another workflow triggered, and would be able to run this directly in the E2E workflows

theodur (changes requested):

.github/workflows/e2e_android.yml
```diff
@@ -9,6 +9,9 @@ on:
     branches:
       - 'dependabot/**'
   workflow_dispatch:
+  workflow_run:
+    workflows: ['[TestRail] Start RC Run']
```
theodur, line 13: If this needs to run whenever a release build is made, shouldn't this be `[Build] Release Build`?

.github/workflows/e2e_ios.yml
```diff
@@ -9,6 +9,9 @@ on:
     branches:
       - 'dependabot/**'
   workflow_dispatch:
+  workflow_run:
+    workflows: ['[TestRail] Start RC Run']
```
theodur, line 13: Same question for this

.github/workflows/e2e_android.yml
rbontrager: I've updated the scripts based on your additional comments.

theodur (approved):
Updates look good!

Sparowhawk (approved):

PR #7199, opened 2023-11-03
feature/6939-Dylan-e2eAndroidSlack

Sparowhawk:
## Description of Change
turn on slack thread and add daily e2e builds

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
turn on slack thread and add daily e2e builds

## PR Checklist

## For QA


.github/workflows/e2e_android.yml
```diff
@@ -33,19 +33,19 @@ env:
...
 
 jobs:
-  # start_slack_thread:
-  #   name: Start Slack thread
-  #   uses: ./.github/workflows/start_slack_thread.yml
-  #   secrets: inherit
-  #   with:
-  #     channel_name: va-mobile-build-alerts
-  #     message: 'E2E Android build process starting. Please see :thread: for results. This process may take a while.'
+  start_slack_thread:
```
theodur, line 39: Since this workflow runs on dependabot PRs now, we'll need to add logic to only have this run when the trigger is a cron:
``` yml
if: github.event_name == 'schedule'
```

.github/workflows/e2e_android.yml
```diff
@@ -109,27 +109,27 @@ jobs:
...
-      #   id: post_failure
-      #   if: ${{!failure()}}
-      #   run: |
-      #     ts=$(curl -X POST -H 'Authorization: Bearer '"$SLACK_API_TOKEN"' ' \
-      #       -H 'Content-type: application/json' \
-      #       --data '{"channel":"'${{needs.start_slack_thread.outputs.channel_id}}'","thread_ts":"'${{needs.start_slack_thread.outputs.thread_ts}}'","text":"Success! All E2E tests have passed"}' \
-      #       https://slack.com/api/chat.postMessage|
-      #       jq -r '.ts')
-      #     echo SLACK_THREAD_TS=${ts} >> $GITHUB_OUTPUT
+      - name: Inform Slack of Failure
```
theodur, line 112: Same with this step, this should only run when the trigger is the cron

.github/workflows/e2e_android.yml
```diff
@@ -109,27 +109,27 @@ jobs:
...
+        if: failure()
+        run: |
+          ts=$(curl -X POST -H 'Authorization: Bearer '"$SLACK_API_TOKEN"' ' \
+            -H 'Content-type: application/json' \
+            --data '{"channel":"'${{needs.start_slack_thread.outputs.channel_id}}'","thread_ts":"'${{needs.start_slack_thread.outputs.thread_ts}}'","text":"Failed! Please check workflow for results: https://github.com/department-of-veterans-affairs/va-mobile-app/actions"}' \
+            https://slack.com/api/chat.postMessage|
+            jq -r '.ts')
+          echo SLACK_THREAD_TS=${ts} >> $GITHUB_OUTPUT
+
+      - name: Inform Slack of success
```
theodur, line 123: Same here also

theodur (changes requested):
Added some comments on when we should prevent starting a Slack thread.

Also, I noticed this will create a separate thread for Android tests, and the [sibling PR](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/7200) will do the same for the iOS tests. It would be nice if we could have a single thread for both iOS and Android tests, similar to how we do with on demand builds. That way, the results would be in one place. This would probably require a new workflow for running both test suites, because they'd need to reference the same Slack thread

.github/workflows/e2e_android.yml
Sparowhawk: added

.github/workflows/e2e_android.yml
Sparowhawk: added

.github/workflows/e2e_android.yml
```diff
@@ -36,19 +36,20 @@ env:
...
-  #     channel_name: va-mobile-build-alerts
-  #     message: 'E2E Android build process starting. Please see :thread: for results. This process may take a while.'
+  start_slack_thread:
+    if: github.event_name == 'schedule'
+    name: Start Slack thread
+    uses: ./.github/workflows/start_slack_thread.yml
+    secrets: inherit
+    with:
+      channel_name: va-mobile-build-alerts
+      message: 'E2E Android build process starting. Please see :thread: for results. This process may take a while.'
```
theodur, line 49: I think including something in the message about running tests would make the intent of the thread more clear, as it doesn't mention anything about actually running the tests. Maybe something like
```
Starting E2E Android tests. Please see :thread: for results. This process may take a while.
```

That also brings up another small concern that's beyond the scope of this ticket. I noticed this is being sent to the `va-mobile-build-alerts` channel. This channel has been primarily used for build alerts, as suggested in the channel name. If we're going to start adding different types of alerts to this channel, I wonder if it'd make sense to rename the channel to represent a broader range of alerts. Since E2E tests are still somewhat related to builds (i.e. they tell us when something would be wrong with a build), maybe they still belong with build alerts, but just something for us to think about moving forward

Sparowhawk: updated, I see the point of renaming the channel at some point, but at the moment it still has a descriptive enough name

theodur (approved):

alexandec (approved):

rbontrager (approved):

PR #7222, opened 2023-11-07
CU/6035-Dylan-RemoveOldE2EPackages

Sparowhawk:
## Description of Change
We are no longer using the old method of running E2E tests so we should remove appium, chromedriver, and other related packages from our package.json

## Screenshots/Video
e2e tests still run

Screenshot 2023-11-07 at 9 54 58 AM
Screenshot 2023-11-07 at 11 41 04 AM

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
We are no longer using the old method of running E2E tests so we should remove appium, chromedriver, and other related packages from our package.json

detox should still run

## PR Checklist

## For QA


theodur (commented):
You can also remove the [wdio group](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/df1f4c007f8c14063863e478dde8d88de9259dfc/.github/dependabot.yml#L35-L37) in the dependabot config. There's also a merge conflict, but you'll need to sync with `develop` anyway because it looks like your branch is behind since it doesn't have the dependabot config

Sparowhawk:
> You can also remove the [wdio group](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/df1f4c007f8c14063863e478dde8d88de9259dfc/.github/dependabot.yml#L35-L37) in the dependabot config. There's also a merge conflict, but you'll need to sync with `develop` anyway because it looks like your branch is behind since it doesn't have the dependabot config

updated

Sparowhawk:
just wanted to note that when we finish up the component unit tests we can remove the react-test-render dependencies

when the component library is done taking all the components we should be able to remove the styled-components and jest-styled-components libraries

RN-fetch-blob  we might be able to turn into a native module since it is only used in one file

reduxjs/toolkit might be able to be removed once we refactor everything to react query

Might be able to remove some others in the future as we make changes as well. Haven't investigate each one thoroughly to see how much work it would be. These just came off the top of my head based on code I know we are actively changing.

alexandec:
Ran some E2E tests on this branch locally and they worked fine

alexandec (approved):

dumathane (approved):

rbontrager (approved):

PR #7230, opened 2023-11-08
bug/7229-theo-ClearSIDOnLogout

theodur:
## Description of Change

Currently, when a user logs out of the app and a different user logs into the app on the same device, push notifications for the previously logged-in user are still sent to the new user's session. ~~This PR fixes that by resetting the device endpoint SID on logout, so that the next time a user logs in, the device will register with the push service again.~~

This PR stores the user's ID in async storage after device registration, and compares the stored user ID to the user ID of the new user logged in. If the ID's match, registration will happen, otherwise registration is skipped since it's the same user.

## Screenshots/Video

N/A

## Testing

Testing was completed by the VEText team on a staging build.

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/store/slices/authSlice.ts
```diff
@@ -521,6 +521,9 @@ export const logout = (): AppThunk => async (dispatch, getState) => {
     await clearStoredAuthCreds()
     api.setAccessToken(undefined)
     api.setRefreshToken(undefined)
+    // Remove SID to trigger device registration with push service on next login. This ensures
+    // future logins won't receive push notifications for the user that was previously logged-in.
+    await AsyncStorage.removeItem(DEVICE_ENDPOINT_SID)
```
Sparowhawk, line 526: doesn't this just unregisters the device locally?

Sure the new user would be newly registered with this device, but wouldn't the old user still be registered with this device since we pass those old details to the BE with /v0/push/register. We don't clear what the BE has when someone logs out is what I am getting at and that is the primary thing that controls push notifications reaching the device.

Sparowhawk (approved):

alexandec (approved):

TKDickson (commented):
Dropping my DM questions in this PR as well for documentation sake:

* For a single user on a device, is this changing PN behavior (previously they would have received PN even if they had logged out on their device, now they won't)?
* Doesn't this impact more than just explicit logout, because of https://github.com/department-of-veterans-affairs/va-mobile-app/blob/976431b274c557924182116a93761a1dbbbffb66/VAMobile/src/store/api/api.ts#L155?

TKDickson:
@dumathane has asked us to not move forward with this change; we should be checking with Product/UX about the change in behavior

dumathane:
I feel like there is a solution we could propose that includes a user identifier that is the cleanest way forward. In registerDevice we should save some unique user identifier along with the deviceToken and the SID. On login/auth we should confirm the logging in user is still the same, otherwise we should re-register for push notifications. That way we can fix the push notification <-> user issue while also only bothering to re-register when we actively need to. @theodur let me know if that makes sense and we can confirm with Shane.

dumathane (changes requested):
Tie device token and endpoint sid to a user.

theodur:
@dumathane That makes sense to me. We'll just need to determine what the unique identifier for the user should be. Doesn't look like we get anything back from the [auth service](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/6f962e16c0b695eeb6cf428389d69a4ebf6f45a5/VAMobile/src/store/api/types/auth.ts#L31-L42) that we can use. I'll bring this up with Shane tomorrow

dumathane:
image
Maybe id here?

dumathane:
Chatted with Kris and this is a user uuid and could probably serve as a unique enough identifier.

timwright12:
@theodur any movement on this one?

theodur:
@timwright12 @dumathane Just saw Binny's comments right now since I was OOO on Friday. I think the `id` field from the user endpoint makes sense. One potential issue I see is we probably won't have the `id` field at the time of registering the device, but I guess we can hit the user endpoint at the time of registration. I'll look into that today

dumathane (approved):

Sparowhawk (approved):

alexandec (approved):

TKDickson:
Android isn't allowing opt-in for PN on this branch; you can on develop. Details in ticket.

TKDickson (approved):
I've done basic regression testing, and VEText team has told Theo they've done the functionality testing on this ticket. Ready to merge to develop.

PR #7252, opened 2023-11-13
Attempted app update button

dumathane:
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/components/WaygateWrapper.tsx
+++ b/VAMobile/src/components/WaygateWrapper.tsx
@@ -1,4 +1,4 @@
-import { AlertBox, Box, ClickToCallPhoneNumber } from 'components'
+import { AlertBox, Box, ButtonTypesConstants, ClickToCallPhoneNumber, VAButton } from 'components'
 import { useNavigationState } from '@react-navigation/native'
 import React, { FC } from 'react'
 
@@ -6,6 +6,7 @@ import { NAMESPACE } from 'constants/namespaces'
 import { Waygate, WaygateToggleType, waygateEnabled } from 'utils/waygateConfig'
 import { a11yLabelID } from 'utils/a11yLabel'
 import { displayedTextPhoneNumber } from 'utils/formattingUtils'
+import { openAppStore } from 'utils/homeScreenAlerts'
 import { useTheme } from 'utils/hooks'
 import { useTranslation } from 'react-i18next'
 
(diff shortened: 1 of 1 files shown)


alexandec (approved):

VAMobile/src/components/WaygateWrapper.tsx
```diff
@@ -24,7 +25,10 @@ export const WaygateWrapper: FC<WaygateWrapperProps> = ({ children, waygateName
     return (
       <Box mb={theme.dimensions.condensedMarginBetween}>
         <AlertBox border={waygate.type === 'DenyContent' ? 'error' : 'warning'} title={waygate.errorMsgTitle} text={waygate.errorMsgBody}>
-          <ClickToCallPhoneNumber displayedText={displayedTextPhoneNumber(t('8006982411'))} phone={t('8006982411')} a11yLabel={a11yLabelID(t('8006982411'))} />
+          <Box my={theme.dimensions.standardMarginBetween}>
```
TimRoe, line 28: Should this also be conditional on waygate.appUpdateButton? Otherwise it doesn't match up to existing behavior--there'll be a bunch of extra padding on top/bottom.

Also: maybe this should just be `mb`? Not `my`.

VAMobile/src/components/WaygateWrapper.tsx
```diff
@@ -24,7 +25,10 @@ export const WaygateWrapper: FC<WaygateWrapperProps> = ({ children, waygateName
...
+            <ClickToCallPhoneNumber displayedText={displayedTextPhoneNumber(t('8006982411'))} phone={t('8006982411')} a11yLabel={a11yLabelID(t('8006982411'))} />
+          </Box>
+          {waygate.appUpdateButton === true && <VAButton onPress={openAppStore} label={t('encourageUpdate.update')} buttonType={ButtonTypesConstants.buttonPrimary} />}
```
TimRoe, line 31: Nitpicky, but isn't this the same as:
```suggestion
          {waygate.appUpdateButton && <VAButton onPress={openAppStore} label={t('encourageUpdate.update')} buttonType={ButtonTypesConstants.buttonPrimary} />}
```

TimRoe, line 31: Why is the waygate using `encourageUpdate.update`? Seems like bridging it cross features we should rename the translation to just `update`.

TimRoe (approved):
Approving, but really think we should rename the `encourageUpdate.update` translation to just `update` to reflect it now no longer being tied exclusively to that feature.

VAMobile/src/components/WaygateWrapper.tsx
TimRoe: Binny noted via Slack this was intentional.

TimRoe (approved):

Issue #5859, opened 2023-06-02
CU - Replace FormWrapper test cases with the new testing system


dumathane:
## Proposed Change

We bypassed some troublesome test cases recently on [this](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/5699) PR, and will need to revisit the FormWrapper test cases given their new functionality.

## Why Should We Prioritize?

Shouldn't retire tests that still serve a good purpose.

## Coding Time Estimation

5	

## Testing Considerations

Should only include test cases.

## Checklist
[ ] Add the front-end label tag

[ ] Attach to the Frontend epic for engineering initiatives

PR #7389, opened 2023-11-27
CU/7095-Liz-unit-tests-components-part5

liztownd:
## Description of Change
Updated tests to use best practices:
S src/components/FormWrapper/FormFields/FormAttachments.test.tsx
M src/components/FormWrapper/FormFields/RadioGroup.test.tsx
M src/components/FormWrapper/FormWrapper.test.tsx
S src/components/VAIcon/VAIcon.test.tsx

## Screenshots/Video
No screenshots, unit test updates only.

## Testing
- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations
Using best practices for testing, removed all `testInstance` references and utilized the `screen.getBy*` methods.

## PR Checklist

## For QA


VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
```diff
@@ -51,55 +42,45 @@ context('FormAttachments', () => {
...
 
-  it('initializes correctly', async () => {
-    expect(component).toBeTruthy()
-  })
-
   describe('when there are attachments', () => {
-    it('should display a remove link', async () => {
-      expect(testInstance.findAllByType(TextView)[2].props.children).toEqual('Remove')
+    it('should display a remove link', () => {
+      expect(screen.getAllByText('Remove').length).toBeGreaterThan(0)
```
Sparowhawk, line 60: should be able to update this to getByRole('link', {name: 'Remove'})

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
```diff
@@ -51,55 +42,45 @@ context('FormAttachments', () => {
...
     })
 
     describe('when the remove link is clicked for an attachment', () => {
-      it('should call the removeOnPress', async () => {
-        await waitFor(() => {
-          testInstance.findAllByProps({ accessibilityLabel: 'Remove' })[0].props.onPress()
-          expect(removeOnPressSpy).toHaveBeenCalled()
-        })
+      it('should call the removeOnPress', () => {
+        fireEvent.press(screen.getAllByText('Remove')[0])
```
Sparowhawk, line 65: should be able to update this to getByRole('link', {name: 'Remove'})

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
```diff
@@ -51,55 +42,45 @@ context('FormAttachments', () => {
...
     })
   })
 
   describe('when there are no attachments', () => {
-    it('should not display a remove link', async () => {
+    it('should not display a remove link', () => {
       initializeTestInstance([])
-      expect(testInstance.findAllByType(TextView)[0].props.children).toEqual('Attachments')
-      expect(testInstance.findAllByType(TextView)[1].props.children).toEqual('add files')
+      expect(screen.queryByText('Remove')).toBeFalsy()
```
Sparowhawk, line 74: should be able to update this to queryByRole('link', {name: 'Remove'})

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
```diff
@@ -51,55 +42,45 @@ context('FormAttachments', () => {
...
+      expect(screen.getByText('add files')).toBeTruthy()
```
Sparowhawk, line 75: update this one to the getByRole you used below

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-    it('should set the error to empty string', async () => {
-      await waitFor(() => {
-        let shortenedFieldsList = formFieldsList[1]
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VAModalPicker).props.setError('')
-        expect(screen.queryByText('second error message')).toBeFalsy()
-      })
+    it('should set the error to empty string', () => {
+      initializeTestInstance(pickerFieldList)
+      screen.UNSAFE_getByType(VAModalPicker).props.setError('')
```
Sparowhawk, line 81: there should be another way to get the VAModalPicker instead of this unsafe way that we are trying to get rid of

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-    it('should set the error to empty string', async () => {
-      await waitFor(() => {
-        let shortenedFieldsList = formFieldsList[0]
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VATextInput).props.setError('')
-        expect(screen.queryByText('first error message')).toBeFalsy()
-      })
+    it('should set the error to empty string', () => {
+      initializeTestInstance(textInputFieldList)
+      screen.UNSAFE_getByType(VATextInput).props.setError('')
```
Sparowhawk, line 89: similar to the other one, there should be another way that isn't this unsafe way to get this

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-    it('should set the error to empty string', async () => {
-      await waitFor(() => {
-        let shortenedFieldsList = formFieldsList[2]
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VASelector).props.setError('')
-        expect(screen.queryByText('third error message')).toBeFalsy()
-      })
+    it('should set the error to empty string', () => {
+      initializeTestInstance(selectorFieldList)
+      screen.UNSAFE_getByType(VASelector).props.setError('')
```
Sparowhawk, line 97: similar to the other one, there should be another way that isn't this unsafe way to get this

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-    it('should set the error to the field error message', async () => {
-      await waitFor(() => {
-        let shortenedFieldsList = formFieldsList[1]
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VAModalPicker).props.setError()
-        expect(screen.getByText('second error message')).toBeTruthy()
-      })
+    it('should set the error to the field error message', () => {
+      initializeTestInstance(pickerFieldList)
+      screen.UNSAFE_getByType(VAModalPicker).props.setError()
```
Sparowhawk, line 105: similar to the other one, there should be another way that isn't this unsafe way to get this

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-    it('should set the error to the field error message', async () => {
-      await waitFor(() => {
-        let shortenedFieldsList = formFieldsList[0]
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VATextInput).props.setError()
-        expect(screen.getByText('first error message')).toBeTruthy()
-      })
+    it('should set the error to the field error message', () => {
+      initializeTestInstance(textInputFieldList)
+      screen.UNSAFE_getByType(VATextInput).props.setError()
```
Sparowhawk, line 113: similar to the other one, there should be another way that isn't this unsafe way to get this

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-    it('should set the error to the field error message', async () => {
-      await waitFor(() => {
-        let shortenedFieldsList = formFieldsList[2]
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VASelector).props.setError()
-        expect(screen.getByText('third error message')).toBeTruthy()
-      })
+    it('should set the error to the field error message', () => {
+      initializeTestInstance(selectorFieldList)
+      screen.UNSAFE_getByType(VASelector).props.setError()
```
Sparowhawk, line 121: similar to the other one, there should be another way that isn't this unsafe way to get this

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -63,112 +59,85 @@ context('FormWrapper', () => {
...
-        initializeTestInstance([shortenedFieldsList])
-        testInstance.findByType(VASelector).props.setError()
-        expect(screen.getByText('third error message')).toBeTruthy()
-
-        initializeTestInstance([shortenedFieldsList], true)
-        expect(screen.getByText('I confirm that this information is correct. (Required)')).toBeTruthy()
-      })
+    it('should clear the errors object', () => {
+      initializeTestInstance(selectorFieldList)
+      screen.UNSAFE_getByType(VASelector).props.setError()
```
Sparowhawk, line 129: similar to the other one, there should be another way that isn't this unsafe way to get this

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
```diff
@@ -185,70 +154,7 @@ context('FormWrapper', () => {
           expect(onSaveSpy).toHaveBeenCalled()
         })
       })
```
Sparowhawk, line 156: I believe there is another ticket in the backlog #5859 to update these unit tests below to be able to function. So please take a look and see if there are any other tests that you can possibly test again or things that may be able to be tested. Once you have done so feel free to close out 5859 as this ticket should cover that update

VAMobile/src/components/VAIcon/VAIcon.test.tsx
```diff
@@ -1,11 +1,5 @@
...
 import React from 'react'
-// Note: test renderer must be required after react-native.
-import 'jest-styled-components'
-import { ReactTestInstance, act } from 'react-test-renderer'
-
-import { context, render, RenderAPI, waitFor } from 'testUtils'
-import HomeSelected from './navIcon/homeSelected.svg'
+import { context, render, screen } from 'testUtils'
 import VAIcon from './VAIcon'
 import { InitialState } from 'store/slices'
```
Sparowhawk, line 4: fairly certain initialState can be removed like your part 4 ticket

VAMobile/src/components/VAIcon/VAIcon.test.tsx
```diff
@@ -57,18 +29,11 @@ context('VAIconTests', () => {
...
-        Object.assign(
-          {},
-          {
-            height: 3,
-            width: 6,
-          },
-        ),
-      )
+      expect(screen.getByTestId('myId')).toBeTruthy()
+      expect(screen.UNSAFE_getByProps({
```
Sparowhawk, line 33: this unsafe test can be removed as we aren't testing styling with these changes

Sparowhawk (changes requested):

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
liztownd: I can use `getAllByRole` but there's still more than one, so it has to use the "all" bit. Updated.

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
liztownd: Same as above, it has to use the "all" modifier because there are more than one. Updated.

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
liztownd: Updated.

VAMobile/src/components/FormWrapper/FormFields/FormAttachments.test.tsx
liztownd: Updated.

VAMobile/src/components/VAIcon/VAIcon.test.tsx
liztownd: Updated

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
liztownd: Reading through the docs the reason those methods are called `UNSAFE_` is only in reference to how you have to have implicit knowledge of the props and types that are being passed, which we do here, that is the point, we've got to access that prop to test the error handling, so I'm not convinced that in this instance we should try to get rid of them...

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
liztownd: See above.

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
liztownd: See above.

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
liztownd: See above.

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
liztownd: See above.

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
liztownd: won't fix.

Sparowhawk (approved):

VAMobile/src/components/FormWrapper/FormWrapper.test.tsx
alexandec: Looking at React Testing Library's [Guiding Principles](https://testing-library.com/docs/guiding-principles/):

> If it relates to rendering components, then it should deal with DOM nodes rather than component instances, and it should not encourage dealing with component instances.

RTL's general philosophy is to test components as a user would see/interact with them, and to avoid testing implementation details whenever possible. The more implementation details you interact with, the more fragile the tests become because refactoring can break the tests even if the functionality is still complete and correct from the user's perspective.

I feel like these types of tests, "when the picker calls setError with an empty string" it "should set the error to empty string" are somewhat in the weeds of implementation details. Is there a way the user would interact with this component that would test these types of scenarios in a more complete way, without digging into the implementation details? If so that would be more in line with the RTL approach.

liztownd: I get that, and I agree, but from what I understand (granted, I might be wrong) the Wrapper basically just wraps the form fields passed into it and does validation and error handling for the form.... I'm happy to remove all the "sets the correct errors" tests, but that doesn't leave much _to_ test. 

As for how the user interacts with it, the user doesn't necessarily interact directly with this component wrapper, only the child form components (except maybe the test for the "reset errors")... @Sparowhawk and I spoke about it and agreed that while these are `UNSAFE_` methods, and yes, we're getting a little too into the weeds with testing, if we want test coverage on this component, we need to test the functionality, and this was the only way I could see to do that. :( 

I'm certainly open to suggestions if you want to poke at it and come up with a better way to test it though!!

alexandec: I looked at this some more, and I think the best approach would be to set the child form components to have invalid or empty values, and simulate the saved state for the form. That should cause FormWrapper to show the error messages. The final test in this file, just before the commented-out stuff, has some of the right approach. It's not exactly right since it is setting props directly, but it does simulate the state after onSave is called

alexandec (approved):

Issue #6999, opened 2023-10-13
Make it easier to sign into demo mode


TKDickson:
## Description 
As a user of the VA app demo mode, I want to log in quickly, so that I can get to the good stuff faster.

As part of this story we want to:   
Request from Ryan Thurlwell in DSVA slack:

> If, by chance, we can make it easier to sign into demo mode when we make these updates, that would be amazing. 21 taps + password with upper/lower/special is annoying. As an example, the State Farm app just uses a 4 character user name, and no password at all and it’s so nice to access quickly and on the fly.

Note other requirements of the above comment are captured in this issue:
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7410

## Acceptance Criteria
  

-  [x] lower the number of logo taps required to show the password prompt

**Accessibility Requirements**
- [ ] AC 1

## Notes & Open Questions

- Dependencies/Roadblocks:
    - Any internal/external dependencies?
    - Test accounts needed?
- Does this require QA?
    
- Dev Notes: 

## Ticket Checklist

- [ ] Acceptance criteria defined
- [ ] Labels added (front-end, back-end, feature)
- [ ] Linked to an Epic

rbontrager:
Verified that the number of taps to get into demo mode has decreased to 7 and that the change has caused no issues with our detox tests.  Approved by QA.

PR #7399, opened 2023-11-27
#6999 Less Taps To Sign Into Demo + Timeout

hasahmed2:
## Description of Change

*NOTE* I am a new developer and this is my first PR. All guidance is appreciated

Reduced the number of logo taps required to get the prompt for demo mode to 7, and also added a timeout so that the taps must all happen in quick succession.

## Screenshots/Video

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/148818278/d36f982a-7ff2-4835-a27e-5434367f1b7e

## Testing

- [x] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


Sparowhawk (approved):

alexandec:
@hasahmed2 could you please link the relevant ticket using the "Connect issue" button?

alexandec:
@hasahmed2 it's this button here, hopefully it's visible to you:

Screenshot 2023-11-29 at 10 24 17 AM

If you can't see it, that might be because you need to be signed into ZenHub

VAMobile/src/screens/auth/LoginScreen/LoginScreen.tsx
```diff
@@ -34,8 +34,10 @@ const LoginScreen: FC = () => {
   const startAuth = useStartAuth()
   const theme = useTheme()
   const [demoPromptVisible, setDemoPromptVisible] = useState(false)
-  const TAPS_FOR_DEMO = 20
+  const TAPS_FOR_DEMO = 7
+  const DEMO_TAPS_TIMEOUT_MS = 500
```
alexandec, line 38: Are there any a11y concerns with adding this timeout to access demo mode? My concern is that someone who can't tap as fast might hit the timeout and be unable to access demo mode. Also is adding this timeout a requirement mentioned in the ticket?

alexandec (changes requested):

VAMobile/src/screens/auth/LoginScreen/LoginScreen.tsx
hasahmed2: No requirement in the ticket for timeout. As for a11y for timeout, I'm not sure, is there someone I can ask about that?

As for timeout, my thinking was that we don't want people accidentally triggering the demo-mode prompt, but that was purely an assumption on my part that that may be a desired way to handle it. It can be easily removed or duration could be extended. Which would be preferred?

dumathane: I would just drop this, and I feel comfortable making that call solo without leverage accessibility etc.

alexandec (approved):
Looks good, thanks for fixing that number of taps comparison too

alexandec:
@hasahmed2 this still needs to be linked to the ticket using the "Connect issue" button (see screenshot above)

rbontrager (approved):

PR #7443, opened 2023-12-04
bug/5034-chris-claims-file-upload-font-size-fix

alexandec:
## Description of Change

This PR:
- Adds a `maxWidth` prop to VAIcon which checks the scaled width and limits scaling to respect maxWidth
- Uses the new prop to limit icon size in PhotoAdd and PhotoPreview components
- Limits text scaling in the same two components

The new maxWidth prop should be useful throughout the app in cases where we need to limit icon scaling.

## Screenshots/Video

Before/after (iPhone):
&nbsp;&nbsp;

Before/after (Android):
&nbsp;&nbsp;

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

Claim file upload icons and text should scale up to an appropriate max size, even at the largest accessibility size settings.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/components/VAIcon/VAIcon.tsx
+++ b/VAMobile/src/components/VAIcon/VAIcon.tsx
@@ -1,6 +1,5 @@
 import { AppState, AppStateStatus } from 'react-native'
 import { SvgProps } from 'react-native-svg'
-import { isFinite } from 'underscore'
 import React, { FC, useEffect } from 'react'
 
 import { AccessibilityState } from 'store/slices'
@@ -168,6 +167,9 @@ export type VAIconProps = BoxProps & {
   /**  optional number use to set the height; otherwise defaults to svg's height */
   height?: number
 
+  /** optional maximum width when scaled */
+  maxWidth?: number
+
   /** optional boolean that prevents the icon from being scaled when set to true */
   preventScaling?: boolean
 
@@ -193,7 +195,7 @@ export type VAIconProps = BoxProps & {
  *
  * @returns VAIcon component
  */
-const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, preventScaling, testID, ...boxProps }) => {
+const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, maxWidth, preventScaling, testID, ...boxProps }) => {
   const theme = useTheme()
   const fs: (val: number) => number = useFontScale()
   const dispatch = useAppDispatch()
@@ -223,12 +225,32 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, pre
     return <></>
   }
 
-  if (width && isFinite(width)) {
-    iconProps = Object.assign({}, iconProps, { width: preventScaling ? width : fs(width) })
+  // Default to scaled dimensions
+  const scaledWidth = width && fs(width)
+  const scaledHeight = height && fs(height)
+  let finalWidth = scaledWidth
+  let finalHeight = scaledHeight
+
+  if (preventScaling) {
+    // Use un-scaled dimensions
+    finalWidth = width
+    finalHeight = height
+  } else if (scaledWidth && maxWidth) {
+    // If scaled width > max, use max
+    if (scaledWidth > maxWidth) {
+      finalWidth = maxWidth
+      if (height) {
+        // Scale height using same ratio as width
+        const scalingRatio = finalWidth / width
+        finalHeight = scalingRatio * height
+      }
+    }
   }
 
-  if (height && isFinite(height)) {
-    iconProps = Object.assign({}, iconProps, { height: preventScaling ? height : fs(height) })
+  iconProps = {
+    ...iconProps,
+    ...(finalWidth && { width: finalWidth }),
+    ...(finalHeight && { height: finalHeight }),
   }
 
   return (
--- a/VAMobile/src/components/VAIcon/VAIcon.tsx
+++ b/VAMobile/src/components/VAIcon/VAIcon.tsx
@@ -1,6 +1,5 @@
 import { AppState, AppStateStatus } from 'react-native'
 import { SvgProps } from 'react-native-svg'
-import { isFinite } from 'underscore'
 import React, { FC, useEffect } from 'react'
 
 import { AccessibilityState } from 'store/slices'
@@ -226,7 +225,7 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, max
     return <></>
   }
 
-  if (width && isFinite(width) && height && isFinite(height)) {
+  if (width && height) {
     // Default to scaled dimensions
     const scaledWidth = fs(width)
     let finalWidth = scaledWidth
--- a/VAMobile/src/components/VAIcon/VAIcon.tsx
+++ b/VAMobile/src/components/VAIcon/VAIcon.tsx
@@ -1,5 +1,6 @@
 import { AppState, AppStateStatus } from 'react-native'
 import { SvgProps } from 'react-native-svg'
+import { isFinite } from 'underscore'
 import React, { FC, useEffect } from 'react'
 
 import { AccessibilityState } from 'store/slices'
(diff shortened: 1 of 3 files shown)


TimRoe:
Did we do anything like this elsewhere in the app? Maybe a question for design, but feels weird that I think every other screen would reflow the previews to be each on their own line (and correspondingly larger) while this one maintains 3 columns in the face of large scaling.

VAMobile/src/components/VAIcon/VAIcon.tsx
```diff
@@ -223,12 +225,32 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, pre
...
+  const scaledWidth = width && fs(width)
+  const scaledHeight = height && fs(height)
+  let finalWidth = scaledWidth
+  let finalHeight = scaledHeight
+
+  if (preventScaling) {
+    // Use un-scaled dimensions
+    finalWidth = width
+    finalHeight = height
+  } else if (scaledWidth && maxWidth) {
```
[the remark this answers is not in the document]
TimRoe: Nvm, guess that's the way it was before too.

VAMobile/src/components/VAIcon/VAIcon.tsx
```diff
@@ -223,12 +225,32 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, pre
     return <></>
   }
 
-  if (width && isFinite(width)) {
-    iconProps = Object.assign({}, iconProps, { width: preventScaling ? width : fs(width) })
+  // Default to scaled dimensions
+  const scaledWidth = width && fs(width)
+  const scaledHeight = height && fs(height)
```
TimRoe, line 230: Doesn't need to exist as a const, only used once.

VAMobile/src/components/VAIcon/VAIcon.tsx
```diff
@@ -223,12 +225,32 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, pre
     return <></>
   }
 
-  if (width && isFinite(width)) {
```
TimRoe, line 226: While I like removing the if width/height checks in theory to reduce nesting, in practice I think it makes the code more  brittle since we're doing manipulation that only makes sense if they exist and effectively checking if they exist multiple times now with the scaled___ consts and then again with the final___ to actually override in iconProps. 

Coming back to this later seems more likely to cause a bug than it just being clear from the outset that the current structure does not make sense to scale or max scale unless width/height are explicitly set (since we're not reading the values from the SVG size).

TimRoe (changes requested):
Dropping a request changes just to be fully clear between my general comment vs. code change comments:
1. I think we should verify with design this is the direction we should resolve this bug vs adjusting the "preview" structure to reflow onto separate lines which would make the icon sizes fine and I think better fit other workflows at large scaling
2. If design is cool with this way of resolving, think it could use some reworking as I think it has become more brittle than what existed before

TimRoe:
> I'm not certain of the larger implications of reducing the number of columns. Maybe it would depend on how important it is for the user to see each image at an extra large size, vs the importance of seeing multiple images without scrolling as much?

If they're using very high scaling, my assumption is they have considerable challenge seeing generally and have 0 concerns about how much scrolling it causes. But would defer to design as it is previews of what they previously selected.

> We limit the indicator size in claims here: https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/src/utils/claims.tsx#L267-L269.

Hah, thought there was one spot. However, the way that works where the circle can contain a number or icon is a bit of a mess that maybe could be improved if it was refreshed sometime down the road.

alexandec:
@TimRoe I looked into the background on this a bit more and found [a ticket](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/6488) where Brea reviewed possible solutions. Her conclusion was:

> The acceptance criteria have been updated with two possible options to fix the issue. My preference is for option 1, which recommends setting max sizes.

She's referring to the two options in the [ticket for this PR](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/5034) which are:

Option 1 & preference: We keep the 3-column layout, but set max sizes for:
- Thumbnail images
- Icons within the thumbnails (set max size or to a percentage within the thumbnail)
- The thumbnail label (file size and/or "add photo" copy)

Option 2:
- We figure out a way to allow each of these elements to expand (similar to grid / flexbox) and wrap to help accommodate the increased font size.

Based on Brea's preference for Option 1 I'll keep the basic approach here, which adds a maxWidth.

alexandec:
@TimRoe see what you think of this restructured version. It checks for width and height up front

VAMobile/src/components/VAIcon/VAIcon.tsx
```diff
@@ -223,12 +226,24 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, pre
...
-    iconProps = Object.assign({}, iconProps, { width: preventScaling ? width : fs(width) })
-  }
+  if (width && isFinite(width) && height && isFinite(height)) {
```
TimRoe, line 229: 1. Any idea why isFinite was being used? At face value removing it seemed fine
2. This condition doesn't work correctly as is--does nothing if you only tried to set width or height and not both (which I assume breaks existing behavior somewhere)

I think we might be trying to overcomplicate this, couldn't we just:
```
const truncateWidth = width && fs(width) > maxWidth
// Mostly same as old code
if (width) {
  iconProps = { ...iconProps, width: preventScaling ? width : truncateWidth ? maxWidth : fs(width) }}
}

if (height && truncateWidth) {
  const scaledHeight = ( maxWidth / width ) * height
  iconProps = { ...iconProps, { height: preventScaling ? height : scaledHeight }}
} else if (height) {
  iconProps = { ...iconProps, { height: preventScaling ? height : fs(height) }}
}
```

Maybe could be cleaned up more, but fundamentally: the only real special handling needed is the height in the case we've truncated width.

VAMobile/src/components/VAIcon/VAIcon.tsx
alexandec: Removed

VAMobile/src/components/VAIcon/VAIcon.tsx
alexandec: I looked through our existing uses of VAIcon, and found only two cases:
- Both width and height specified
- No dimensions specified

I tried specifying only width or only height to VAIcon to see what happens, and it's not good. The icon file itself specifies the missing dimension so you get an awkward mix of what you specified on one dimension and what the icon file specified on the other dimension. From what I can tell we don't ever actually want that. In order to correctly calculate the missing dimension we'd have to read the file to determine the aspect ratio.

I don't think the isFinite check is needed either. I assume it was added originally to ensure people didn't try to specify values like '22px' but since we have a TS checking the type, that should be sufficient to warn people that they didn't provide a number. I removed the isFinite check.

VAMobile/src/components/VAIcon/VAIcon.tsx
```diff
@@ -223,12 +225,24 @@ const VAIcon: FC<VAIconProps> = ({ name, width, height, fill, fill2, stroke, pre
     return <></>
   }
 
-  if (width && isFinite(width)) {
-    iconProps = Object.assign({}, iconProps, { width: preventScaling ? width : fs(width) })
-  }
-
-  if (height && isFinite(height)) {
-    iconProps = Object.assign({}, iconProps, { height: preventScaling ? height : fs(height) })
+  if (width && height) {
```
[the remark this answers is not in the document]
TimRoe: Hah, and you committed a cleaned up way as soon as I added this. I like that too.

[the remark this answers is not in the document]
alexandec: Just restructured to make it simpler, see what you think of this iteration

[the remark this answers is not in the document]
alexandec: Haha cool 👍

Sparowhawk (approved):

theodur (approved):

DJUltraTom (approved):

PR #7481, opened 2023-12-06
Enhancement/5264 - Hasan Ahmed - Swap Large List Nav for List Rows on Personal Information Screen

hasahmed2:
## Description of Change

The Personal Information Page was using Large List Nav components but that was the wrong component for use-case. So this story swaps it for DefaultList to match with other similar screens

## Screenshots/Video

image&nbsp;&nbsp;image

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


Sparowhawk (approved):

dumathane (approved):
Approving this. In the future Hasan, can you attach a before and an after pic? Also, your SS from android can be taken in the emulator with the little snapshot button so it will match in relative size to the iOS version.

PR #7486, opened 2023-12-06
feature/5534-Dylan-UpdateReactNative

Sparowhawk:
## Description of Change
updated RN to 72.7(73 had a lot of release candidates and didn't know which one they would go with)
Also had to update lottie-react-native to 5.1.6 to take care of a conflict which also allowed us to drop the patch for it

## Screenshots/Video
N/A

## Testing
yarn test and sniff tests on both iOS and Android

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
upgrade RN to latest

## PR Checklist

## For QA


dumathane (approved):

alexandec (dismissed):

Sparowhawk:
/Users/runner/work/va-mobile-app/va-mobile-app/VAMobile/ios/Pods/Headers/Public/Flipper-Folly/folly/portability/Config.h:20:10: 'folly/folly-config.h' file not found

#include <folly/folly-config.h>

This is the current error in the testrail run for iOS

I have tried the following:

Full clean (remove node modules, remove pods, remove ios/build folder, cleared derived data, and rebuilt from scratch)
latest flipper folly, 0.164, 0.163, and the current 0.125 versions
reverted entire podfile to what we currently have
reverted podfile with the different versions of flipper folly
tried adding things to gitignore
also tried -DFOLLY_NO_CONFIG adding this flag

The issue is this error above is preventing the archiving of the app in CI. It has something to do with the podfile, RN version, flipper folly, and xcode. It may be the RN version is to new for the xcode version that is used by CI.

alexandec:
Locally with Xcode 15.1 I get a folly error unless I do the following:

- Select Pods > Build Settings > In section Apple Clang - Preprocessing > under Macro section
- Add in release & debug `_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION`

I wonder if something similar is possible on CI

VAMobile/ios/Podfile
```diff
@@ -1,5 +1,9 @@
-require_relative '../node_modules/react-native/scripts/react_native_pods'
-require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+# Resolve react_native_pods.rb with node to allow for hoisting
+require Pod::Executable.execute_command('node', ['-p',
+  'require.resolve(
+    "react-native/scripts/react_native_pods.rb",
```
TimRoe, line 4: Given this is now directly using a ruby file, did we need to add Ruby in the Gemfile as the upgrade helper has?

VAMobile/ios/VAMobile/AppDelegate.mm
```diff
@@ -58,16 +57,6 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
 
-/// This method controls whether the `concurrentRoot`feature of React18 is turned on or off.
-///
-/// @see: https://reactjs.org/blog/2022/03/29/react-v18.html
-/// @note: This requires to be rendering on Fabric (i.e. on the New Architecture).
-/// @return: `true` if the `concurrentRoot` feature is enabled. Otherwise, it returns `false`.
-- (BOOL)concurrentRootEnabled
-{
-  // Switch this bool to turn on and off the concurrent root
-  return false;
```
TimRoe, line 69: Should this have been removed? I see it was in the upgrade helper, but also we had it disabled and [searching on it the doc indicates it's an opt out](https://reactnative.dev/docs/react-18-and-react-native#opt-ing-out-of-react-18-on-ios). Not sure if it's trying to run with it on without it being explicitly false and that could be causing issues.

Sparowhawk:
I tried removing flipper entirely from iOS, but it is actually RCT/Folly that is the issue

alexandec (approved):

DJUltraTom (approved):

timwright12 (approved):

PR #7494, opened 2023-12-07
cu/7101-chris-unit-tests-components-part10

alexandec:
## Description of Change

Improve component unit tests, part 10. It's the last unit test improvement ticket! 🎉 

Also remove a few unneeded things now that the tests are all switched over.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

Successful unit test run. Tests follow best practices

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/testUtils.tsx
+++ b/VAMobile/src/testUtils.tsx
@@ -2,13 +2,12 @@ import { I18nextProvider } from 'react-i18next'
 import { Provider } from 'react-redux'
 import { render as rtlRender } from '@testing-library/react-native'
 import { ThemeProvider } from 'styled-components'
-import React, { ElementType } from 'react'
+import React from 'react'
 import i18nReal from 'utils/i18n'
 import { RootState } from 'store'
 import path from 'path'
 import { AnyAction, configureStore, Store } from '@reduxjs/toolkit'
 import { NavigationContainer } from '@react-navigation/native'
-import { ReactTestInstance } from 'react-test-renderer'
 import { QueryClient, QueryClientProvider, QueryKey } from '@tanstack/react-query'
 
 import accessabilityReducer from 'store/slices/accessibilitySlice'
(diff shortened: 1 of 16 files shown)


Sparowhawk (approved):

VAMobile/src/testUtils.tsx
```diff
@@ -2,13 +2,12 @@ import { I18nextProvider } from 'react-i18next'
...
 import { render as rtlRender } from '@testing-library/react-native'
 import { ThemeProvider } from 'styled-components'
-import React, { ElementType } from 'react'
+import React from 'react'
 import i18nReal from 'utils/i18n'
 import { RootState } from 'store'
 import path from 'path'
 import { AnyAction, configureStore, Store } from '@reduxjs/toolkit'
 import { NavigationContainer } from '@react-navigation/native'
-import { ReactTestInstance } from 'react-test-renderer'
```
Sparowhawk, line 11: can't we also remove this import and the @types associated with it in the package.json now as well?

Sparowhawk (changes requested):

VAMobile/src/testUtils.tsx
Sparowhawk: can you also check to see if @types/styled-components-react-native can be removed as well?

alexandec: I had also hoped we could remove react-test-renderer since we don't import it anywhere now, but unfortunately removing it produces this error when running a test:

```
  ● Test suite failed to run

    Cannot find module 'react-test-renderer' from 'node_modules/@testing-library/react-native/build/act.js'
```

so I guess react native native testing library uses it internally.

@types/styled-components-react-native can't be removed either. Removing it produces a bunch of errors like:

```
[1] src/components/FooterButton.tsx(12,20): error TS7016: Could not find a declaration file for module 'styled-components'. '/Users/chris/Documents/code/va-mobile-app/VAMobile/node_modules/styled-components/dist/styled-components.cjs.js' implicitly has an 'any' type.
```

due to usage of styled-components inside some of our components

Sparowhawk (approved):

theodur (approved):

alexandec:
Changes to tests only, no QA needed. Will merge

Issue #7485, opened 2023-12-06
Apply Use Focus Effect to Home Screen


ajsarkar28:
AC:

- When a user focuses on the home screen, all Personalization APIs (Secure Message, Appointments, Claims & Appeals, and Prescriptions) should be reloaded

TKDickson:
@theodur things I'm noticing, testing with karsyn:
* We're running the appointments and Rx APIs on the sync screen, and then running them again immediately once the home screen loads. Is there a way to move/remove API calls to get rid of that duplication?
* The claims call for Karsyn is NOT running on the sync screen, but is running (and failing) on the home screen. I'm not sure which of those is correct -- are we not calling claims when we should on the sync screen, or are we incorrectly calling it on the home screen? Karysn is authorized for appeals, but not claims, according to authorizedServices.
* After the home screen loads, there's a significant delay before any action I take will be reflected in the app (~6 seconds on Android, and ~42 seconds on iOS). It's probably the thing I've seen before, where we get some errors back immediately from the backend but it freezes the front end for several seconds, but that's very concerning to happen on the home screen, especially since the problem is opaque to veterans.

TKDickson:
Noticing, testing with +54:
* Claims isn't reloading/resending when I go back to the home screen. The API fires the first time I'm on the home screen, and again when I go to the claims list, but not at each HS load.

theodur:
@TKDickson Nice finds; to answer your questions:

> We're running the appointments and Rx APIs on the sync screen, and then running them again immediately once the home screen loads. Is there a way to move/remove API calls to get rid of that duplication?

Binny has a [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7300) he's working on for removing API calls from the sync screen, but that won't include appointments and RX, since his ticket will be based off and merged into `develop`. I can remove the appointment and RX API calls on the sync screen in my branch to follow that trend. Definitely not ideal to have successive calls like that.

> The claims call for Karsyn is NOT running on the sync screen, but is running (and failing) on the home screen. I'm not sure which of those is correct -- are we not calling claims when we should on the sync screen, or are we incorrectly calling it on the home screen? Karysn is authorized for appeals, but not claims, according to authorizedServices.

After looking at the code, it doesn't look like there's a check if the user is authorized for the service before making the API call, which is probably the issue. This is also the case for the other API calls on the home screen. I'll need to add a check for all of those API calls.

[...]

PR #7495, opened 2023-12-08
feature/7379-theo-HomeScreenAppointmentsModule

theodur:
## Description of Change

Adds logic for displaying the upcoming appointments module on the home screen. The appointments module is displayed if there are upcoming appointments when the user first signs in and views the home screen.

**Note**: The current UI is just a skeleton for the logic, since the UI will be covered in a separate ticket. There's also a separate [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7485) for handling reloading the appointments count when the home screen is focused.

## Screenshots/Video

Screenshot 2023-12-11 at 9 07 02 AM

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HomeScreen/HomeScreen.tsx
+++ b/VAMobile/src/screens/HomeScreen/HomeScreen.tsx
@@ -5,7 +5,7 @@ import { useTranslation } from 'react-i18next'
 import React, { FC } from 'react'
 
 import { AppointmentsState } from 'store/slices'
-import { Box, CategoryLanding, EncourageUpdateAlert, Nametag, SimpleList, SimpleListItemObj, TextView, VAIconProps } from 'components'
+import { Box, CategoryLanding, EncourageUpdateAlert, LargeNavButton, Nametag, SimpleList, SimpleListItemObj, TextView, VAIconProps } from 'components'
 import { CloseSnackbarOnNavigation } from 'constants/common'
 import { Events } from 'constants/analytics'
 import { FEATURE_LANDING_TEMPLATE_OPTIONS } from 'constants/screens'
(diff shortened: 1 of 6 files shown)


VAMobile/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -76,6 +81,16 @@ export const HomeScreen: FC<HomeScreenProps> = ({ navigation }) => {
       <Box flex={1} justifyContent="flex-start">
         <EncourageUpdateAlert />
         <Nametag />
+        {upcomingAppointmentsCount && (
+          <Box mx={theme.dimensions.gutter} mb={theme.dimensions.condensedMarginBetween}>
+            <LargeNavButton
+              title={`${t('appointments')}`}
+              subText={`(${upcomingAppointmentsCount} ${t('upcoming')})`}
+              onPress={() => Linking.openURL('vamobile://appointments')}
```
theodur, line 89: Unfortunately, the appointments screen needs to be opened via a deep link to ensure back navigation is handled properly. This will most likely be the case with the other home screen modules.

Sparowhawk (approved):

dumathane (approved):

TKDickson:
@theodur seems like an appointmentCount of 0 is crashing the app during login -- has happened on at least two users in staging (most recently, kidys.b.quankep@id.me).

theodur:
@TKDickson The issue with the app crashing should be fixed now, just pushed a fix to the branch

dumathane:
Should we be changing Appointments.tsx to not unconditionally make api calls since we are now doing it earlier in the process?
  
  useEffect(() => {
    const todaysDate = DateTime.local()
    const threeMonthsEarlier = todaysDate.minus({ months: 3 })

    const upcomingRange: AppointmentsDateRange = getUpcomingAppointmentDateRange()
    const pastRange: AppointmentsDateRange = {
      startDate: threeMonthsEarlier.startOf('day').toISO(),
      endDate: todaysDate.minus({ days: 1 }).endOf('day').toISO(),
    }

    // fetch upcoming and default past appointments ranges
    if (apptsNotInDowntime) {
      dispatch(prefetchAppointments(upcomingRange, pastRange, ScreenIDTypesConstants.APPOINTMENTS_SCREEN_ID))
    }
  }, [dispatch, apptsNotInDowntime])

theodur:
@dumathane Yeah, `prefetchAppointments` uses previously fetched appointments when available, so we should be good in this case. But I agree, definitely something to keep an eye out for other modules.

Initially I was considering making the `past` parameter in the function optional, and only fetching upcoming appointments on the sync screen. This would remove an API call since the function calls the API twice (once for upcoming and once for past appointments), but I decided to leave it for now since past appointments would need to be fetched anyway if the user tapped the appointments module and launched to the appointments screen. This is something we can reconsider if we ever need to improve the performance of the sync screen's API calls down the line

TKDickson (approved):
All cases pass, ready to merge to XL

PR #7591, opened 2023-12-19
bug/7590-rachael-detox-github_actions_run_testRail

rbontrager:
## Description of Change

It was determined that the test-rail integration script was never going to run do to existing github actions logic (https://github.com/actions/runner/issues/2205).  The test-rail integration script was changed so it always runs and the testRail.yml was slightly changed so it will create a new test run if necessary. 
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


.github/workflows/e2e_android.yml
```diff
@@ -119,6 +110,13 @@ jobs:
           avd-name: Pixel_4_XL_API_28
           script: yarn e2e:android-test --take-screenshots failing --updateSnapshot --headless
 
+      - name: Upload e2e-junit
+        if: failure() || success()
```
Sparowhawk, line 114: I don't believe you need the if statement here as it should work no matter what

.github/workflows/e2e_android.yml
```diff
@@ -148,15 +146,9 @@ jobs:
...
-      - name: Upload e2e-junit
-        uses: actions/upload-artifact@v3
-        with:
-          name: e2e-junit
-          path: VAMobile/e2e/test_reports/e2e-junit.xml
-      
   send_test_results_to_testrail:
-    if: (github.event_name == 'workflow_run' && '!cancelled()' && github.event.workflow_run.conclusion != 'skipped') || ${{ github.event.inputs.run_testRail == true}}
-    needs: [e2e-android]
+    if: ${{ always() }}
```
Sparowhawk, line 150: I don't think this if statement needs to be there

.github/workflows/e2e_ios.yml
```diff
@@ -106,6 +97,13 @@ jobs:
       - name: Run e2e tests for iOS
         run: yarn e2e:ios-test --take-screenshots failing --updateSnapshot
 
+      - name: Upload e2e-junit
+        if: failure() || success()
```
Sparowhawk, line 101: same with this if statement

.github/workflows/e2e_ios.yml
```diff
@@ -134,16 +132,10 @@ jobs:
...
-      - name: Upload e2e-junit
-        uses: actions/upload-artifact@v3
-        with:
-          name: e2e-junit
-          path: VAMobile/e2e/test_reports/e2e-junit.xml
 
   send_test_results_to_testrail:
-    if: (github.event_name == 'workflow_run' && '!cancelled()' && github.event.workflow_run.conclusion != 'skipped') || ${{ github.event.inputs.run_testRail == true}}
-    needs: [e2e-ios]
+    if: ${{ always() }}
```
Sparowhawk, line 137: same with this if statement

Sparowhawk (changes requested):

.github/workflows/e2e_ios.yml
rbontrager: This is necessary. If you take a look at the link I have in the description, there is currently some github actions logic that will cause it not to run this job without that statement when start_slack_job is skipped (ie when we run it manually).

Sparowhawk (dismissed):

.github/workflows/e2e_android.yml
```diff
@@ -148,15 +155,9 @@ jobs:
...
-        uses: actions/upload-artifact@v3
-        with:
-          name: e2e-junit
-          path: VAMobile/e2e/test_reports/e2e-junit.xml
-      
   send_test_results_to_testrail:
-    if: (github.event_name == 'workflow_run' && '!cancelled()' && github.event.workflow_run.conclusion != 'skipped') || ${{ github.event.inputs.run_testRail == true}}
-    needs: [e2e-android]
+    if: always() && (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion != 'skipped') || github.event.inputs.run_testRail == 'true'
+    needs: e2e-android
```
theodur, line 160:
```suggestion
    needs: e2e-android
```

.github/workflows/e2e_ios.yml
```diff
@@ -9,17 +9,8 @@ on:
     branches:
       - 'dependabot/**'
   workflow_dispatch:
-    inputs:
-      run_testRail:
-        description: Run_TestRail
-        type: boolean
-        required: false
```
theodur, line 16: Looks like this was accidentally removed?

.github/workflows/e2e_ios.yml
```diff
@@ -9,17 +9,8 @@ on:
...
   schedule:
     - cron: '0 4 * * 1,2,3,4,5'
-  workflow_run:
-    workflows: ['[Release] New Release Issue']
-    types: 
-      - completed
```
theodur, line 22: Same with this trigger

rbontrager: Added back in

.github/workflows/e2e_ios.yml
rbontrager: Added back in

theodur (approved):

PR #7681, opened 2024-01-10
bug/7402-chris-vama-claim-count-revision

alexandec:
## Description of Change

Modify `vama_claim_count` event to remove tab param, and send event on screen load after claims request completes.

## Screenshots/Video

Console output when visiting the Claims History page:

```
logging analytics event vama_claim_count {"closed_claims": 1, "open_claims": 7}
```

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

- Modify vama_claim_count to only fire when a user navigates to the screen ClaimsHistory.
- Continue to capture the custom parameters closed_claims and open_claims
- Remove claims_tab parameter.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/store/slices/claimsAndAppealsSlice.ts
+++ b/VAMobile/src/store/slices/claimsAndAppealsSlice.ts
@@ -48,6 +48,7 @@ export type ClaimsAndAppealsState = {
   loadingAppeal: boolean
   loadingSubmitClaimDecision: boolean
   loadingFileUpload: boolean
+  finishedLoadingClaimsAndAppeals: boolean
   error?: Error
   claimsServiceError: boolean
   appealsServiceError: boolean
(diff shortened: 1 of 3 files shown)


VAMobile/src/store/slices/claimsAndAppealsSlice.ts
```diff
@@ -48,6 +48,7 @@ export type ClaimsAndAppealsState = {
   loadingAppeal: boolean
   loadingSubmitClaimDecision: boolean
   loadingFileUpload: boolean
+  finishedLoadingClaimsAndAppeals: boolean
```
alexandec, line 51: Had to add this because the `loadingClaimsAndAppeals` value is initially false, then becomes true, then false again. The initial false value caused the analytics event to be sent immediately on initial render, then sent again after the data finished loading

Sparowhawk (approved):

alexandec:
Nothing to QA on this one since it's just an analytics change. We don't yet have a way for QA to check analytics events

PR #7689, opened 2024-01-11
feature/Binny-7300-Pulling-API-Calls-Forward

dumathane:
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/screens/SyncScreen/SyncScreen.test.tsx
theodur, line 1: Since the new API calls will be removed from the sync screen and moved to the home screen, the changes to this file can be reverted

PR #7708, opened 2024-01-16
feature/7530-theo-HomeScreenModuleDisability

theodur:
## Description of Change

Adds disability ratings and monthly benefit payment to the home screen.

**Note:** The UI is temporary while the design is being finalized.
## Screenshots/Video

image

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -87,6 +91,14 @@ export const HomeScreen: FC<HomeScreenProps> = ({ navigation }) => {
     }, [dispatch, smInDowntime, userAuthorizedServices?.secureMessaging]),
   )
 
+  useFocusEffect(
+    useCallback(() => {
+      if (userAuthorizedServices?.decisionLetters && !lettersInDowntime) {
```
dumathane, line 96: I don't think this authorization is required for this info?

VAMobile/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -156,6 +168,18 @@ export const HomeScreen: FC<HomeScreenProps> = ({ navigation }) => {
...
+            <TextView variant={'MobileBodyBold'}>{t('disabilityRating.title')}</TextView>
+            <TextView>{`${t('disabilityRating.combinePercent', { combinedPercent: ratingData.combinedDisabilityRating })}`}</TextView>
+          </Box>
+        )}
+        {letterBeneficiaryData?.benefitInformation.monthlyAwardAmount && (
+          <Box mx={theme.dimensions.gutter} mb={theme.dimensions.condensedMarginBetween}>
+            <TextView variant={'MobileBodyBold'}>{t('monthlyPayment')}</TextView>
+            <TextView>{`$${roundToHundredthsPlace(letterBeneficiaryData.benefitInformation.monthlyAwardAmount)}`}</TextView>
+          </Box>
+        )}
```
dumathane, line 182: https://adhoc.slack.com/archives/C02N65QRRB7/p1702415611135249

Should we be avoiding this type of code as a best practice for this threads reasons?

dumathane (changes requested):
Just looking for responses/changes to my two comments.

VAMobile/src/screens/HomeScreen/HomeScreen.tsx
theodur: Hmm initially I though it would either be `decisionLetters` or `lettersAndDocuments` required for it. I noticed the only other place we're calling it is [here](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/4022d1a21156c113128e92ab90bfbcaa385217a9/VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx#L57), and we're not even checking any authorized service 🤔. I'll do some digging and see

VAMobile/src/screens/HomeScreen/HomeScreen.tsx
theodur: Strange because it doesn't seem like it's been an issue with other cases in the past. Even the React [documentation](https://react.dev/learn/conditional-rendering#logical-and-operator-) suggests using the `&&` for conditionally rendering content. I think it's only an issue when the left side of the conditional is the number `0`, since that'll cause issues, which is why I checked if the number `activeClaimsCount` is greater than 0 for rendering, that way the left side will always be a boolean. I actually missed that for the `combinedDisabilityRating` rendering logic though. I'll go ahead and update that one

dumathane (approved):

TKDickson (approved):
ready to merge to XL

PR #7791, opened 2024-01-24
CU/6920-roettger-DesignSystemEngineeringDoc

TimRoe:
## Description of Change

- Filled out the "For engineers" to be an entire section of documentation instead of a single page
- Updated the Docusaurus local building doc to add a new step
- Added a VAMobile 1Pass vault item containing the auth key for pulling the engineering doc over from the `va-mobile-library` repo
   - A temporary measure until that repo is made public after which that auth key, the 1Pass item, and the Docusaurus process change should be reverted
   - This was also added as a repo secret under `AUTH_DOC_SITE`

General note: the content from the `va-mobile-library` repo is pulled in every time the Doc site is built so the automated daily run (or a manual run) will pull in updates made to the doc there.

## Screenshots/Video

N/A

## Testing

Checking things worked locally building Docusaurus.

## PR Checklist

## For QA


.github/workflows/documentation_deploy.yml
```diff
@@ -14,6 +14,8 @@ jobs:
   deploy:
     name: Deploy documentation site to GitHub Pages
     runs-on: ubuntu-latest
+    env:
+      AUTH_DOC_SITE: ${{ secrets.AUTH_DOC_SITE }}
```
TimRoe, line 18:
```suggestion

```

PR #7870, opened 2024-01-31
feature/5336-rachael-detox-military-history-expansion

rbontrager:
## Description of Change

Add ability to test multiple branches of service

## Screenshots/Video

Screenshot 2024-01-31 at 9 37 15 AM

Screenshot 2024-01-31 at 9 36 57 AM

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


alexandec:
I saw a failure running these locally:

```
  ● Military Info Screen › [ios] should show correct information if no military service is available

    Test Failed: Failed expectation: TOEXIST WITH MATCHER(label == “We're sorry. We can't access your military service records. If you think you should be able to review your service information here, please file a request to change or correct your D-D 2 1 4 or other military records.”)
```

The full set of tests took about 15 minutes to complete since it appears they're reinstalling the app after each service change. I assume there's no other way to do it, but it definitely slows things down significantly

rbontrager:
@alexandec The failure looks to be another case where the detox tests were failing when the env variable is_test is set to true locally and not failing when is_test is set to false and was not failing on github actions.  With that being said I changed to look for the id that I stuck in nomilitaryinformationaccess and didn't have any failures when is_test was true and false locally. In regards to the app reinstalling, I tried everything I can think of not to have the app reinstall but the mock data wasn't updating.  I was planning on only using this when we have no other option but to change the json.

alexandec (approved):

PR #7881, opened 2024-02-01
feature/6227-Chika-AddUpdateAnalyticsLogin

cadibemma:
## Description of Change

- Adds new event `vama_login_readme` and updates use of events `vama_login_start` and `vama_login_success`

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

- [ ] Confirm that these events aren't firing anywhere else

## PR Checklist

## For QA


VAMobile/src/constants/analytics.ts
```diff
@@ -572,11 +572,20 @@ export const Events = {
       },
     }
   },
-  vama_login_start: (isSIS = false): Event => {
+  vama_login_readme: (isSIS = false): Event => {
```
Sparowhawk, line 575: SIS is the only available login method now so I don't think this is required anymore and everything should be defaulted to true now

VAMobile/src/constants/analytics.ts
```diff
@@ -572,11 +572,20 @@ export const Events = {
...
+    return {
+      name: 'vama_login_readme',
+      params: {
+        sis: isSIS.toString(),
+      },
+    }
+  },
+  vama_login_start: (isSIS = false, isBiometric = false): Event => {
```
Sparowhawk, line 583: same thing here. we haven't used the old login version in years and is no longer supported and can be hardcoded to true or dropped all together I believe

cadibemma: I defaulted sis here to true

VAMobile/src/constants/analytics.ts
cadibemma: I decided to remove sis since I though it may not be necessary for the readme event

VAMobile/src/components/CollapsibleView.tsx
```diff
@@ -33,10 +37,14 @@ export type CollapsibleViewProps = {
  */
 const CollapsibleView: FC<CollapsibleViewProps> = ({ text, contentInTextArea = true, showInTextArea = true, a11yHint, textColor, children, testID }) => {
   const theme = useTheme()
+  const { t } = useTranslation(NAMESPACE.COMMON)
   const [expanded, setExpanded] = useState(false)
 
   const onPress = (): void => {
     setExpanded(!expanded)
+    if (!expanded && text === t('loaGate.expandMsg')) {
```
dumathane, line 45: Ya, we can do better than this. We need to log all of these with a parameter not just a single one if we're going to keep using this component. I'll follow up with you.

dumathane (dismissed):
Will talk to you.

VAMobile/src/constants/analytics.ts
theodur: I'd prefer dropping the `isSIS` param altogether to avoid confusion since it's not relevant anymore, unless there's a reason we still want to keep it?

VAMobile/src/store/api/types/Screens.ts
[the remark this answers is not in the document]
cadibemma: Ok! I removed it from the list

theodur (approved):

PR #7887, opened 2024-02-01
feature/7885-Dylan-UpdateAFContentCapabilities

Sparowhawk:
## Description of Change
Updated waygates to have a v2 string that takes into account these special characters
It also supplies a a11yLabel to handle any VA potential spots
\n Newline
\r Carriage Return
\t Tab
\u2022 Bullet Points

Also allowed the override of what phone number is supplied to the alert box

Update

dropped support for the following as it wasn't supported on some Android devices:
\u2028 Line Separator
\u2029 Paragraph Separator
\v Vertical Tab

## Screenshots/Video

Screenshot 2024-02-01 at 5 28 46 PM

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Code a way to include string manipulation (whitespace, carriage return, etc.) that doesn't break backwards compatibility.
 Include a new variable for overriding the displayed phone number in ClickToCallPhoneNumber, default to the previous existing.
 Ensure all accessibility concerns are addressed for the new approach.
 All additions need to be included in the developer menu remote config objects.

## PR Checklist

## For QA


VAMobile/src/utils/jsonFormatting.ts
```diff
@@ -0,0 +1,25 @@
...
+  \r Carriage Return
+  \u2028 Line Separator
+  \u2029 Paragraph Separator
+  \t Tab
+  \v Vertical Tab
+  \u2022 Bullet Points
+  @param text - array of textline to concatenate
+  */
+
+export const fixedWhiteSpaceString = (text: string | undefined) => {
```
alexandec, line 13: Do we expect this will be used in other places in the future? Just checking whether it needs to be in its own utils file, or can just be placed where it's used in the waygate stuff

alexandec (dismissed):

dumathane (changes requested):
I'd like video uploads of android and ios screenreaders correctly reading 'VA' and then still having access to the phone numbers.

VAMobile/src/utils/jsonFormatting.ts
Sparowhawk: @alexandec this needs to be in it's own file as I have to ignore the file in the lint config. It doesn't allow double backslashes in the rules apparently even though that is the proper way to escape a back slash for regex.

alexandec: @Sparowhawk you can disable eslint for a single line, or for a block of code. Docs here: https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1. I would recommend that rather than creating a separate file and adding it to the eslint config.

Sparowhawk:
https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/850657e7-2bad-41e7-a669-268f6561c326

Sparowhawk:
https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/caa482fe-40d2-4e1a-85f0-62856c6acb96

VAMobile/src/components/WaygateWrapper.tsx
```diff
@@ -40,17 +41,20 @@ export const WaygateWrapper: FC<WaygateWrapperProps> = ({ children, waygateName,
...
+    const text = errorMsgBodyV2.length > 0 ? errorMsgBodyV2 : waygate.errorMsgBody
+    const phoneNumber = waygate.errorPhoneNumber && waygate.errorPhoneNumber.length > 0 ? waygate.errorPhoneNumber : t('8006982411')
     return (
       <Box mb={theme.dimensions.condensedMarginBetween}>
         <AlertBox
           border={waygate.type === 'DenyContent' ? 'error' : 'warning'}
           title={waygate.errorMsgTitle}
-          text={waygate.errorMsgBody}
+          text={text}
+          textA11yLabel={a11yLabelVA(text || '')}
```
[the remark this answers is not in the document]
Sparowhawk: Uploaded videos for both, it works on both

Sparowhawk:
> I'd like video uploads of android and ios screenreaders correctly reading 'VA' and then still having access to the phone numbers.

uploaded for both and it worked for both ios and android

VAMobile/src/utils/jsonFormatting.ts
Sparowhawk: no I cannot this particular eslint rule is not able to be disabled so the only solution I found was to ignore the file

Sparowhawk:
https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/756921cd-bdc9-4649-9048-72b6e15b8dc1

Sparowhawk:
https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/5422b5be-cfde-4920-9a09-6325a5d6cbf9

dumathane (dismissed):

rbontrager:
@Sparowhawk I've tested this on two android devices I have (Pixel 3 XL running android 12 and samsung galaxy 8 running android 10) and the \u2028 and \u2029 is showing up with the invalid character symbol and \v is not working at all.

dumathane (approved):

rbontrager (approved):

PR #7890, opened 2024-02-02
Update Mobile App Org Chart to reflect new structure 

bischoffa:
## Description of Change

Update Mobile App Org Chart to reflect new structure 

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,5 +14,6 @@ This repository hosts the source code for the VA: Health and Benefits mobile app
 
 See the [team folder](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/products/va-mobile-app) for all the background, discovery, planning, and decisions that preceded application development.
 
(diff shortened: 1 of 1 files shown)


bischoffa:
@TimRoe here is the updated org chart with Victor on it and not Greta. Can we upload this up using this PR ? Or do I need to create another one?

TimRoe:
I think Adam meant @timwright12 ?

Also, at a glance that image you linked looks like it's in `private-user-images` so I don't think it'd be visible to the public which I assume it should be.

bischoffa:
Apologies meant @timwright12   
  
The original image Seth shared. I downloaded it and edited it to add Victor. Not sure if that is what its private user

timwright12:
@bischoffa you can close this out, I just updated it right on the branch

PR #7955, opened 2024-02-08
feature/7263-rachael-detox-appointments-expansion

rbontrager:
## Description of Change

As part of this story we want to make a new test that covers the following for appointments:

Add testing for each upcoming/confirmed appointment type
Add testing for each past/confirmed appointment type
Add testing for each canceled/confirmed appointment type
Add testing for claim exams
Add testing for missing appointment details. Verify that missing details cause the correct descriptions to be displayed

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/.github/workflows/build_ios.yml
+++ b/.github/workflows/build_ios.yml
@@ -91,8 +91,6 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: 'VAMobile/.nvmrc'
-          # cache: 'yarn'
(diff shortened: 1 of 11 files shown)


alexandec:
Succeeds for me locally, but it took almost an hour to complete. Maybe instead of starting on the first page for every appointment, the tests could just keep moving through the list? I feel like that could save half an hour at least

alexandec (approved):

PR #7976, opened 2024-02-13
feature/7909-rachael-detox-github-actions-enable-testing-of-single-test

rbontrager:
## Description of Change

As Qart, I want the ability to to only test specific features in e2e instead of the whole app so that I can gain efficiencies in only running what I / FE needes. 

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


.github/workflows/update_testrail_results.yml
```diff
@@ -86,3 +89,4 @@ jobs:
             --run-id ${{steps.run-id-selection.outputs.TEST_RUN_ID}} \
             --section-id ${{steps.section-id-selection.outputs.SECTION_RUN_ID}} \
             -f "/home/runner/work/va-mobile-app/va-mobile-app/e2e-junit.xml"
+
```
Sparowhawk, line 92: missing new line

Sparowhawk (approved):
Looks good overall. The only thing that I think could be improved is the potential ordering of the steps to potentially reduce the number of steps that repeat

.github/workflows/e2e_android.yml
```diff
@@ -10,6 +10,10 @@ on:
       - 'dependabot/**'
   workflow_dispatch:
     inputs:
+      tests_to_test:
```
alexandec, line 13: Consider calling this "tests to run" instead of "tests to test", I think it's a bit clearer. Same for the other uses below

alexandec (approved):

.github/workflows/e2e_android.yml
```diff
@@ -57,8 +61,91 @@ jobs:
       channel_name: va-mobile-build-alerts
       message: 'Starting E2E Android tests. Please see :thread: for results. This process may take a while.'
 
+  matrix-e2e-android:
```
alexandec, line 64: Since almost all of this seems to be shared with  the existing `e2e-android` job, maybe you could refactor it out to avoid duplication (I think this is similar to what Dylan suggested in his comment)

.github/workflows/update_testrail_results.yml
rbontrager: Added new line

.github/workflows/e2e_android.yml
rbontrager: Changed to tests_to_run

.github/workflows/e2e_android.yml
rbontrager: Dylan and I had a discussion and because refactoring this would probably need either a new reusable workflow or a composite action Dylan and I discussed holding off and doing any refactoring in ticket 7373 (ticket to combine the two workflows).

Sparowhawk (approved):

PR #8088, opened 2024-02-28
CU/6876-Dylan-LetterReactQueryMigration

Sparowhawk:
## Description of Change
migrate to react query, had to do pod install as Theo's revert of linking removed RNFetchBlob

## Screenshots/Video
N/A

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
migrate to react query and functionality stay the same

## PR Checklist

## For QA


VAMobile/src/api/letters/downloadLetter.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return true
+  }
+}
+
+/**
+ * Returns a query for a user letter
+ */
+export const useDownloadLetter = (
+  letterType: LetterTypes,
+  demoMode: boolean,
```
theodur, line 44: Instead of passing `demoMode` as a param, you can directly access the store, and check for it that way

```tsx
store.getState().demo.demoMode
```

VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx
```diff
@@ -50,24 +46,41 @@ type BenefitSummaryServiceVerificationProps = StackScreenProps<
 function BenefitSummaryServiceVerification({ navigation }: BenefitSummaryServiceVerificationProps) {
   const { t } = useTranslation(NAMESPACE.COMMON)
   const theme = useTheme()
-  const dispatch = useAppDispatch()
-  const { downloading, letterBeneficiaryData, mostRecentServices, letterDownloadError, loadingLetterBeneficiaryData } =
-    useSelector<RootState, LettersState>((state) => state.letters)
+  const { demoMode } = useSelector<RootState, DemoState>((state) => state.demo)
```
theodur, line 49: This wouldn't be needed if the state was directly accessed in the query, as mentioned in my earlier comment

theodur (approved):

theodur (commented):
You can also be able to merge changes from `develop` in this to resolve the Pod issue

VAMobile/src/api/letters/downloadLetter.tsx
Sparowhawk: updated

VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx
Sparowhawk: updated

VAMobile/src/screens/BenefitsScreen/Letters/GenericLetter/GenericLetter.tsx
```diff
@@ -1,31 +1,56 @@
...
+    unemployable: letterBeneficiaryData?.benefitInformation.hasIndividualUnemployabilityGranted || false,
+    specialMonthlyCompensation: letterBeneficiaryData?.benefitInformation.hasSpecialMonthlyCompensation || false,
+    adaptedHousing: letterBeneficiaryData?.benefitInformation.hasAdaptedHousing || false,
+    chapter35Eligibility: false,
+    deathResultOfDisability: letterBeneficiaryData?.benefitInformation.hasDeathResultOfDisability || false,
+    survivorsAward:
+      letterBeneficiaryData?.benefitInformation.hasSurvivorsIndemnityCompensationAward ||
+      letterBeneficiaryData?.benefitInformation.hasSurvivorsPensionAward ||
+      false,
+  }
```
theodur, line 40: Instead of declaring the letter options in multiple places, you can set this as a const directly in the `useDownloadLetter` query, and just grab the letter beneficiary data from the query client using `getQueryData`. 

If the `useDownloadLetter` query is being called, the letter beneficiary data should already be cached. This will definitely be the case with the home screen personalization changes. But if you want to be safe, you could always check if cached data for it exists and just fetch the query with the query client if it doesn't

VAMobile/src/screens/BenefitsScreen/Letters/GenericLetter/GenericLetter.tsx
```diff
@@ -1,31 +1,56 @@
...
+  const {
+    isLoading: downloading,
+    isError: letterDownloadError,
+    refetch: refetchLetter,
+  } = useDownloadLetter(letterType, lettersOptions, { enabled: downloadLetter })
```
theodur, line 45: Same thing from my earlier comment applies here, since we don't want this query to run automatically, `enabled` can just be set to false in the hook, which removes the need for the `downloadLetter` state

VAMobile/ios/Podfile.lock
```diff
@@ -1027,4 +1027,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 71cfabcb52e3b2ad52dad40bf80370529df9a239
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.11.3
```
theodur, line 1030: Looks like the pods and podfile need to be synced with `develop`, this change is specifying an older version than what's on `develop`

theodur (changes requested):

rbontrager (approved):

VAMobile/src/screens/BenefitsScreen/Letters/GenericLetter/GenericLetter.tsx
Sparowhawk: updated

VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx
```diff
@@ -50,24 +43,40 @@ type BenefitSummaryServiceVerificationProps = StackScreenProps<
...
+    survivorsAward:
+      letterBeneficiaryData?.benefitInformation.hasSurvivorsIndemnityCompensationAward ||
+      letterBeneficiaryData?.benefitInformation.hasSurvivorsPensionAward ||
+      false,
+  }
+  const {
+    isLoading: downloading,
+    isError: letterDownloadError,
+    refetch: refetchLetter,
+  } = useDownloadLetter(LetterTypeConstants.benefitSummary, lettersOptions, { enabled: downloadLetter })
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx
```diff
@@ -50,24 +43,39 @@ type BenefitSummaryServiceVerificationProps = StackScreenProps<
...
+    survivorsAward:
+      letterBeneficiaryData?.benefitInformation.hasSurvivorsIndemnityCompensationAward ||
+      letterBeneficiaryData?.benefitInformation.hasSurvivorsPensionAward ||
+      false,
+  }
+  const {
+    isLoading: downloading,
+    isError: letterDownloadError,
+    refetch: refetchLetter,
+  } = useDownloadLetter(LetterTypeConstants.benefitSummary, lettersOptions, { enabled: false })
```
theodur, line 69: Instead of setting this to `false` in each instance of the hook, you can just hard-code it to `false` in the options of the hook itself. Then you wouldn't need to pass `{ enabled: false }` each time you use the hook

VAMobile/src/screens/BenefitsScreen/Letters/GenericLetter/GenericLetter.tsx
theodur: If the default letter options were stored in the hook, there would be a single source of truth for letter options, which would be beneficial in regard to maintenance and accuracy. If more options need to be added in the future, only one place would need to be updated instead of multiple places, which would prevent the default letter options from being out of sync across multiple files. I'm assuming that's why the slice had the default options stored there also. 

Since it's currently only used in two different screens, it's probably not as big of a deal right now, but it potentially could in the future. I'm fine with either path right now though, it can always be revisited in the future if letter options grow, just wanted to call that out.

VAMobile/src/screens/BenefitsScreen/Letters/BenefitSummaryServiceVerification/BenefitSummaryServiceVerification.tsx
Sparowhawk: updated

VAMobile/src/screens/BenefitsScreen/Letters/GenericLetter/GenericLetter.tsx
Sparowhawk: opting to leave it as is right now.

theodur (commented):
Changes look good, just need to remove the Podfile to resolve the merge conflict \

theodur (approved):

rbontrager (approved):

PR #8135, opened 2024-03-07
feature/8081-rachael-detox-automatically-run-on-PR

rbontrager:
## Description of Change

Add the ability to run the relevant detox test automatically on a PR

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


.github/workflows/e2e_android.yml
```diff
@@ -7,7 +7,7 @@ name: '[e2e] Android Detox Test Run'
 on:
   push:
     branches:
-      - 'dependabot/**'
+      - develop
```
Sparowhawk, line 10: is the objective to run the detox tests after it is pushed to develop? That is what this line would do

.github/workflows/e2e_ios.yml
```diff
@@ -7,7 +7,7 @@ name: '[e2e] iOS Detox Test Run'
 on:
   push:
     branches:
-      - 'dependabot/**'
+      - develop
```
Sparowhawk, line 10: is the objective to run the detox tests after it is pushed to develop? That is what this line would do

.github/workflows/e2e_android.yml
Sparowhawk: if that is the objective here is to run after a pr is merged then I think this is good to go

rbontrager: The objective is to have it run whenever someone makes a change to their PR.  I know we talked about it (and I forgot to take the code out) so I think its correct now but correct me if I'm still wrong.

Sparowhawk (approved):

lindluni:
Hello all :wave: 

@lindluni here with the VA GitHub Admin team.

This PR results in your `e2e_android` tests consuming approximately $1,500/day in consumption to run just this one workflow, which is more that we spend on Actions combined across all of our other repositories. A lot of this has to do with not cancelling existing workflows when pushing new commits to existing PR's, thus unnecessarily running duplicate workflows. 

To do this, you would need to add the `concurrency` group key and properly split your workflow to cancel your existing PR runs when pushing new commits.

We'd ask that you revert this change and open an issue in the [https://github.com/department-of-veterans-affairs/github-user-requests](https://github.com/department-of-veterans-affairs/github-user-requests) repository so we can help you add the proper concurrency group to your PR workflow to protect our Actions spend.

Thank you for your help with this.

PR #8142, opened 2024-03-07
feature/7852-Chika-CreateAndImplementBannerComponentForAnnouncements

cadibemma:
## Description of Change

- Added a new banner component that links to the PACT Act URL

#### Before/After: 
&nbsp;&nbsp;

#### Light/Dark: 
&nbsp;&nbsp;
## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/components/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const pressableStyles: ViewStyle = {
+    flexDirection: 'row',
+  }
+
+  return (
+    <Box {...boxProps}>
+      <Pressable
+        style={pressableStyles}
+        onPress={() => Linking.openURL(`https://${link}`)}
```
theodur, line 46: Since we should be storing and referencing links from the `constant.env` file, we probably don't need to prefix the link since the consts will already be the full URL

VAMobile/src/components/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,68 @@
...
+        accessible={true}
+        accessibilityRole={'button'}
```
theodur, line 48:
```suggestion
        accessibilityRole={'link'}
```

VAMobile/src/styles/themes/standardTheme.ts
```diff
@@ -48,6 +52,10 @@ const fontSizes = {
     fontSize: 16,
     lineHeight: 20,
   },
+  AnnouncementBannerTitle: {
+    fontSize: 16,
+    lineHeight: 20,
```
theodur, line 57:
```suggestion
    lineHeight: 22,
```
Based on the design in Figma

VAMobile/src/components/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,68 @@
...
+        testID={title}>
+        <Box flexDirection={'row'} flex={1}>
+          <Box flex={1}>
+            <TextView variant="AnnouncementBannerTitle">{title}</TextView>
+          </Box>
+          <VAIcon
+            width={24}
+            height={24}
```
theodur, line 56:
```suggestion
            width={28}
            height={28}
```
Based on the design in Figma, the width and height should both be `28`

VAMobile/src/components/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const boxProps: BoxProps = {
+    py: theme.dimensions.formMarginBetween,
+    px: theme.dimensions.cardPadding,
+    backgroundColor: theme.colors.buttonBackground.announcementBanner as BackgroundVariant,
+    style: {
+      shadowColor: colors.black,
+      ...Platform.select({
+        ios: {
+          shadowOffset: { width: 0, height: 6 },
+          shadowOpacity: 0.2,
```
theodur, line 28:
```suggestion
          shadowOpacity: 0.4,
```
Based on the design in Figma, the opacity should be 40%. I'd double-check how this looks in the app after updating though, because to me, this already looks pretty close to the design

VAMobile/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -235,6 +280,12 @@ export function HomeScreen({}: HomeScreenProps) {
         <Box mb={theme.dimensions.contentMarginBottom}>
           <SimpleList items={buttonDataList} />
         </Box>
+        <Box mb={theme.dimensions.contentMarginBottom}>
+          <AnnouncementBanner
+            title={t('learnAboutPACT')}
```
theodur, line 285: Since the title has the text `VA.gov` in it, you'll need to set the accessibility label to `a11yLabelVA(t('learnAboutPACT'))` because TalkBack/VoiceOver on certain devices won't pronounce `VA` correctly. You'll probably need to add an accessibility label prop to the component for setting the a11y label

theodur (changes requested):

theodur (commented):
It also looks like this branch was based off Binny's veteran's status card branch. I don't think we want those changes in this PR, especially since they haven't been approved by QA yet, so they should ideally be removed from this PR

VAMobile/src/components/AnnouncementBanner.tsx
cadibemma: 40% opacity makes it look a bit darker than how it looks on Figma, but I'll changed it to 40% to match the design

theodur: Yeah, it could be my eyes, but I also noticed the drop shadow on Android looks a bit darker than the Figma design. Might have to lower that one as well. We can see what design says in visual QA

theodur (approved):

theodur:
Merging this into Binny's nametag PR so we can have one build for all 3 of our PRs

PR #8179, opened 2024-03-12
feature/7877-Chika-ModifyTopNavigationBarShadow

cadibemma:
## Description of Change

- Updated top navigation bar's shadow to reflect the design in the Figma file

#### Before/After: 
&nbsp;&nbsp;

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/components/Templates/HeaderBanner.tsx
```diff
@@ -161,6 +162,18 @@ const HeaderBanner: FC<HeaderBannerProps> = ({
...
+      ...Platform.select({
+        ios: {
+          shadowOffset: { width: 0, height: 6 },
+          shadowOpacity: 0.4,
+        },
+        android: {
+          elevation: 8,
+        },
+      }),
+    },
```
theodur, line 176: A few things I noticed:
1. This will impact all templates, which will affect nearly all screens in the app. This should only be impacting the `CategoryLanding` screens. One option is to add a prop to the `HeaderBanner` that conditionally adds this shadow style, and use that in the `CategoryLanding` template
2. The shadow looks darker than what's in the Figma file

theodur (changes requested):

VAMobile/src/components/Templates/HeaderBanner.tsx
cadibemma: @theodur for #2 , I did notice that as well but just entered as noted in the figma. But for matching sake I'll change opacity to 0.1

cadibemma: fixed!

VAMobile/src/components/Templates/HeaderBanner.tsx
```diff
@@ -161,6 +165,20 @@ const HeaderBanner: FC<HeaderBannerProps> = ({
...
+            ios: {
+              shadowOffset: { width: 0, height: 6 },
+              shadowOpacity: 0.1,
+            },
+            android: {
+              elevation: 8,
+            },
+          }),
+        }
+      : null,
```
theodur, line 181:
```suggestion
      : undefined,
```

theodur (approved):

PR #8254, opened 2024-03-26
CU/8253-Dylan-NotificationRQMigration

Sparowhawk:
## Description of Change
Migrated Notification Slice to RQ

## Screenshots/Video
N/A

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Notifications should still work as original

## PR Checklist

## For QA


Sparowhawk:
the settings detox test is actually broken and needs to be fixed

Sparowhawk:
per @rbontrager a new mapping for the notifications folder will also be needed

timwright12:
@Sparowhawk is this work blocked or is it still "FE-Needs-Review"?

theodur:
@timwright12 This one will be reviewed by us (global) starting today, so I moved it to the code review column. I just need to migrate my changes from #9017 over here

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.test.tsx
```diff
@@ -52,79 +37,125 @@ context('NotificationsSettingsScreen', () => {
...
+        data: {
+          type: 'string',
+          id: 'string',
+          attributes: {
+            preferences: [apptPrefOn],
+          },
+        },
+      }
+      when(api.get as jest.Mock)
+        .calledWith(`/v0/push/prefs/1`)
```
theodur, line 75:
```suggestion
        .calledWith('/v0/push/prefs/1')
```
Template literals aren't needed here

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.test.tsx
```diff
@@ -52,79 +37,125 @@ context('NotificationsSettingsScreen', () => {
...
+        data: {
+          type: 'string',
+          id: 'string',
+          attributes: {
+            preferences: [apptPrefOff],
+          },
+        },
+      }
+      when(api.get as jest.Mock)
+        .calledWith(`/v0/push/prefs/1`)
```
theodur, line 99:
```suggestion
        .calledWith('/v0/push/prefs/1')
```

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.test.tsx
```diff
@@ -52,79 +37,125 @@ context('NotificationsSettingsScreen', () => {
...
```
theodur, line 125:
```suggestion
        .calledWith('/v0/push/prefs/1')
```

VAMobile/src/api/notifications/setPushPref.tsx
```diff
@@ -0,0 +1,47 @@
...
+      const index = pushPreferences.preferences.findIndex((p) => p.preferenceId === preference.preferenceId)
+      const newPrefSetting: PushPreference = {
+        preferenceId: preference.preferenceId,
+        preferenceName: preference.preferenceName,
+        value: !preference.value,
+      }
+      pushPreferences.preferences.splice(index, 1, newPrefSetting)
+      queryClient.setQueryData(notificationKeys.settings, pushPreferences)
+    },
+    onError: async (error) => {
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/setPushPref.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Returns a mutation for updating users push preference
+ */
+export const useSetPushPref = () => {
+  const queryClient = useQueryClient()
+
+  return useMutation({
+    mutationFn: setPushPref,
+    onSuccess: async (data: unknown, preference: PushPreference) => {
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/registerDevice.tsx
```diff
@@ -0,0 +1,78 @@
...
+      setAnalyticsUserProperty(UserAnalytics.vama_uses_notifications(variables.deviceToken ? true : false))
+    },
+    onSuccess: async (response, variables) => {
+      if (response) {
+        await AsyncStorage.setItem(DEVICE_ENDPOINT_SID, response?.data.attributes.endpointSid)
+        variables.deviceToken && (await AsyncStorage.setItem(DEVICE_TOKEN_KEY, variables.deviceToken))
+        variables.userID && (await AsyncStorage.setItem(USER_ID, variables.userID))
+      }
+    },
+    onError: async (error) => {
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.test.tsx
Sparowhawk: updated

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.test.tsx
Sparowhawk: updated

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.test.tsx
Sparowhawk: updated

VAMobile/src/api/notifications/registerDevice.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const deviceName = await getDeviceName()
+    if (!savedToken || savedToken !== registerDeviceParams.deviceToken || !savedSid || isNewUser) {
+      const params: PushRegistration = {
+        deviceName: deviceName,
+        deviceToken: registerDeviceParams.deviceToken,
+        appName: PUSH_APP_NAME,
+        osName: isIOS() ? PushOsName.ios : PushOsName.android,
+        debug: false,
+      }
+      return await put<PushRegistrationResponse>('/v0/push/register', params)
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/loadPushPreferences.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { notificationsEnabled } from 'utils/notifications'
+
+import { notificationKeys } from './queryKeys'
+import { DEVICE_ENDPOINT_SID } from './registerDevice'
+
+/**
+ * Fetch user push preferences
+ */
+const loadPushPreferences = async (): Promise<LoadPushPreferencesData | undefined> => {
+  const systemNotificationsOn = await notificationsEnabled()
```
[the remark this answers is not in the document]
Sparowhawk: pulled into it's own query

VAMobile/src/api/notifications/loadPushPreferences.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const endpoint_sid = await AsyncStorage.getItem(DEVICE_ENDPOINT_SID)
+  if (!endpoint_sid) return
+  const response = await get<GetPushPrefsResponse>(`/v0/push/prefs/${endpoint_sid}`)
+  const previousData = queryClient.getQueryData(notificationKeys.settings) as LoadPushPreferencesData
+  return {
+    preferences: response?.data.attributes.preferences || [],
+    systemNotificationsOn: systemNotificationsOn,
+    deviceToken: previousData?.deviceToken,
+    initialUrl: previousData?.initialUrl,
+    tappedForegroundNotification: previousData?.tappedForegroundNotification || false,
```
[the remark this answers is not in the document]
Sparowhawk: For this one I don't think this would be an issue as the developer screen would mean we are in demo mode and that means the api isn't hit unnecessarily as we intercept it. So I believe these 3 are fine to leave as is

theodur (changes requested):
Requesting changes based on the possibility of the push preferences endpoint being hit multiple times ([comment](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/8254/files#r1689087751))

VAMobile/src/api/notifications/loadPushPreferences.tsx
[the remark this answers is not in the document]
Sparowhawk: updated

Sparowhawk:
> Requesting changes based on the possibility of the push preferences endpoint being hit multiple times ([comment](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/8254/files#r1689087751))

updated

VAMobile/src/api/notifications/loadPushNotification.tsx
theodur, line 1: Not sure that this is the best use of a query. Queries are intended to be used for asynchronous data, using it as storage for global synchronous data is typically advised against because it's an antipattern.

`deviceToken` probably doesn't need to be stored in any type of global state since it's already stored in AsyncStorage, and it's only referenced in the NotificationSettings screen and the Developer screen. It can be pulled directly from AsyncStorage.

`initialUrl` and `tappedForegroundNotification` will need a different solution, but they're both ultimately set in the `NotificationManager` component. One idea is to turn that component into a context provider and make those values accessible throughout the app, since everything is already wrapped inside of it

theodur (changes requested):

VAMobile/src/api/notifications/loadPushNotification.tsx
Sparowhawk: moved deviceToken to be based off Async storage.

I looked into moving it into a context provider and that was more cumbersome and required a large number of changes that I don't think is worth it, when this works as intended with minimal code.

I would be open to including this into some of our other slice code if you feel that would be a better solution than using RQ for those.

theodur: Hmm, I actually think the context provider approach would reduce code, because you'd just need to create the provider in `NotificationManager`, create state variables for `initialUrl` and `tappedForegroundNotification`, and then wrap the provider around the [view](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/d526ed041bf7e44f191ff92f266188bc12bc12d7/VAMobile/src/components/NotificationManager/NotificationManager.tsx#L106) that's returned from the component. From that, you'd be able to reference those values anywhere with the `useContext` hook since we already have the app wrapped in the `NotificationManager`. Then the code for the `useLoadPushNotification` query wouldn't be needed.

I guess the place where you'd have some friction is the `initializeAuth` function, since it references `tappedForegroundNotification`, and isn't directly in a component, so it can't access the context. But the logic that uses `tappedForegroundNotification` can be moved out of that function and placed in the `AuthGuard` where `initializeAuth` is [called](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/5bc38b22282ddc9b722c7bd178d85844df05e18b/VAMobile/src/App.tsx#L271), e.g.

```tsx
const { tappedForegroundNotification } = useContext(NotificationContext)

...

if (!(loggedIn && tappedForegroundNotification)) {
  console.debug('AuthGuard: initializing')
  dispatch(initializeAuth())
}
```

Using Redux would have been my second suggestion for storing these values. The only thing with that is we'd be keeping around a slice just for these two values, which may be a bit overkill, considering they're not updated frequently. Even with that, I'd still choose Redux over RQ in this case, so I'm open to that as well.

theodur (changes requested):
Left a comment on the alternative solutions for the `useLoadPushNotification` query. Also noticed the `PushNotifications` E2E test is failing

VAMobile/src/api/notifications/loadPushNotification.tsx
Sparowhawk: updated to use provider, but causing fork issues if you could look into that

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.tsx
```diff
@@ -22,90 +34,129 @@ import {
...
-          dispatch(setPushPref(pref))
-        },
-      }
-    })
-    return (
-      <Box mt={condensedMarginBetween}>
-        <SimpleList items={prefsItems} />
-      </Box>
-    )
+    if (pushPreferences) {
```
theodur, line 117: Minor suggestion, you could have an `if` statement that returns early if `pushPreferences` doesn't exist instead of wrapping all the rendering logic in an if statement, e.g.

```tsx
if (!pushPreferences) return
```

VAMobile/src/api/notifications/loadPushPreferences.tsx
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Fetch user push preferences
+ */
+const loadPushPreferences = async (): Promise<LoadPushPreferencesData | undefined> => {
+  const endpoint_sid = await AsyncStorage.getItem(DEVICE_ENDPOINT_SID)
+  if (!endpoint_sid) return
+  const response = await get<GetPushPrefsResponse>(`/v0/push/prefs/${endpoint_sid}`)
+  return {
+    preferences: response?.data.attributes.preferences || [],
+  }
```
theodur, line 20: Returning `undefined` when `endpoint_sid` isn't available is causing an error to appear on the Notification settings screen. This is actually from the changes I made in #9017 that I migrated over. Updating the logic to something like this should fix it

```suggestion
  const endpoint_sid = await AsyncStorage.getItem(DEVICE_ENDPOINT_SID)
  let response

  if (endpoint_sid) {
    response = await get<GetPushPrefsResponse>(`/v0/push/prefs/${endpoint_sid}`)
  }

  return {
    preferences: response?.data.attributes.preferences || [],
  }
```

Even though this fixes the error, the other issue is since the `endpoint_sid` doesn't exist for emulators and simulators, this query won't fetch the mock API response. This means on the simulator, the Notification screen will always show the error alert instead of the mock push preferences. We'll need a way around this. Some potential options are:

- Add logic to make the API call for demo mode (not ideal, last resort)
- Mock the `/v0/push/register` endpoint in demo mode so that it returns a mock `endpointSid` that'll be used in the `registerDevice` mutation to set `DEVICE_ENDPOINT_SID`

The latter option should hopefully work option

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.tsx
theodur, line 1: I noticed when you initially don't have system notifications enabled, and you tap the `Open settings` and enable them, and then navigate back to the app, the screen still shows the `Open settings` alert and doesn't show the push preferences. I think the push preferences will need to be refetched on focus when navigating back to the screen from the system settings. I'm seeing this on Android FWIW

theodur (changes requested):
Most of my comments are suggestions changes that don't impact functionality, but there are 2 things I noticed that are impacting functionality, mainly the comments on the endpoint SID, and navigating back to the Notification settings screen from the system settings

VAMobile/src/App.tsx
```diff
@@ -268,18 +269,22 @@ export function AuthGuard() {
...
+    } else if (!loggedIn) {
+      dispatch(initializeAuth())
+      const listener = (event: { url: string }): void => {
+        if (event.url?.startsWith('vamobile://login-success?')) {
+          dispatch(handleTokenCallbackUrl(event.url))
+        }
+      }
+      const sub = Linking.addEventListener('url', listener)
+      return (): void => {
+        sub?.remove()
```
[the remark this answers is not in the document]
Sparowhawk: I found that it was actually causing issue leaving it outside of the block because it was retriggering and causing hang ups on Judy, it is only necessary to have this trigger when logged out

VAMobile/src/api/notifications/loadPushPreferences.tsx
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/loadSystemNotificationSettings.tsx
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/loadSystemNotificationSettings.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { notificationKeys } from './queryKeys'
+
+/**
+ * Fetch user system notification settings
+ */
+const loadSystemNotificationsSettings = async (): Promise<LoadSystemNotificationsData | undefined> => {
+  const systemNotificationsOn = await notificationsEnabled()
+  return {
+    systemNotificationsOn: systemNotificationsOn,
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/queryKeys.ts
```diff
@@ -0,0 +1,4 @@
+export const notificationKeys = {
+  settings: ['settings'] as const,
+  systemNotifications: ['systemNotifications'] as const,
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/registerDevice.tsx
```diff
@@ -0,0 +1,65 @@
...
+      }
+      return put<PushRegistrationResponse>('/v0/push/register', params)
+    }
+  } else {
+    await AsyncStorage.removeItem(DEVICE_TOKEN_KEY)
+  }
+}
+
+/**
+ * Returns a mutation for regestering users device for push notifications
```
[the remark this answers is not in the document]
Sparowhawk: updating

VAMobile/src/api/notifications/registerDevice.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { isErrorObject } from 'utils/common'
+import { getDeviceName } from 'utils/deviceData'
+import { isIOS } from 'utils/platform'
+
+export const DEVICE_TOKEN_KEY = '@store_device_token'
+export const DEVICE_ENDPOINT_SID = '@store_device_endpoint_sid'
+export const USER_ID = '@store_user_id'
+/**
+ * Updates a user's push preference
+ */
```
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/setPushPref.tsx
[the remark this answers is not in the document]
Sparowhawk: updated

VAMobile/src/api/notifications/loadPushPreferences.tsx
Sparowhawk: updated code here

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.tsx
Sparowhawk: updated

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.tsx
Sparowhawk: I added the isFocused check to the push preference query, so it should be working now

VAMobile/src/api/notifications/loadPushPreferences.tsx
Sparowhawk: I didn't need to mock demo mode for this one to work as intended

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.tsx
Sparowhawk: I checked and it unreliably updates it. I found the issue is all the way down in the native module for android and library for iOS. Frankly there isn't anything I can do to either of those to make them update any faster. We could add a timer to refresh the screen after a brief delay. But that also isn't exactly desireable. Or we could kick them back out of the screen when they return to force them to navigate again and give the needed time for the module/library to update.

theodur: This was actually caused by how we're handling refetching the system notification settings when the app is resumed again. When the user navigates back to the app from the system settings, we aren't refetching system notifications. We need to refetch system settings when the screen is resumed from background. I went ahead and added that change to make things easier.

I also added an extra condition to the `usePushPreferences` hook to be disabled if system settings aren't enabled yet, since we don't need to be calling that API if the user doesn't have notifications enabled

VAMobile/src/api/notifications/getSystemNotificationSettings.tsx
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Returns a query for user system notification settings
+ */
+export const useSystemNotificationsSettings = (options?: { enabled?: boolean }) => {
+  return useQuery({
+    ...options,
+    queryKey: notificationKeys.systemSettings,
+    queryFn: () => getSystemNotificationsSettings(),
+    meta: {
+      errorName: 'loadPushPreferences: Service error',
```
theodur, line 27:
```suggestion
      errorName: 'getSystemNotificationsSettings: Failed to retrieve system notification setting',
```

Wrong function was named here, also the error message should be updated since this wouldn't be a service error

VAMobile/src/api/notifications/useUpdatePushPreferences.tsx
theodur, line 1: Noticed the file was misnamed, you probably meant `updatePushPreferences.tsx`

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/NotificationsSettingsScreen/NotificationsSettingsScreen.tsx
Sparowhawk: thanks

VAMobile/src/components/NotificationManager/NotificationManager.tsx
```diff
@@ -1,48 +1,65 @@
...
 const foregroundNotifications: Array<string> = []
 
+interface NotificationContextType {
+  tappedForegroundNotification: boolean
+  initialUrl: string
+  setTappedForegroundNotification: Dispatch<SetStateAction<boolean>>
+  setInitialUrl: Dispatch<SetStateAction<string>>
+}
+
+export const NotificationContext = createContext<NotificationContextType>({
```
theodur, line 22:
```suggestion
const NotificationContext = createContext<NotificationContextType>({
```
Since this isn't imported anywhere else

theodur (approved):
This looks ready for QA now ✅ I noticed a few minor things that should still be updated, but they shouldn't block QA from testing

VAMobile/src/api/notifications/getSystemNotificationSettings.tsx
Sparowhawk: updated

VAMobile/src/components/NotificationManager/NotificationManager.tsx
Sparowhawk: updated

VAMobile/src/api/notifications/useUpdatePushPreferences.tsx
Sparowhawk: updated

theodur (approved):

theodur (approved):

TKDickson:
Sending back for fixes documented in [This comment](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/7954#issuecomment-2334431638) and [this comment.](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8253#issuecomment-2358843856)

TKDickson (changes requested):
The app-in-foreground problem is [still happening, with no change in behavior](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8253#issuecomment-2361318304).
The number of API calls for registration is [improved, but not fully fixed](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/7954#issuecomment-2361314467).

theodur (approved):

TKDickson (approved):
All bugs found during testing are now either fixed, or also present in develop. Detox cases pass. Approved for merge.

dumathane (approved):

PR #8270, opened 2024-03-28
bug/8044-Dylan-SpecialCharactersSecureMessagingFix

Sparowhawk:
## Description of Change
Fixes & " and ' special characters in secure messages

## Screenshots/Video
Screenshot 2024-03-28 at 9 35 55 AM

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
All special characters should render correctly

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/utils/jsonFormatting.ts
+++ b/VAMobile/src/utils/jsonFormatting.ts
@@ -20,7 +20,7 @@ export const fixedWhiteSpaceString = (text: string | undefined) => {
 
 export const fixSpecialCharacters = (text: string): string => {
   return text
-    .replace(/\&amp;/g, '&')
+    .replace(/&amp;/g, '&')
     .replace(/&ldquo;/g, '"')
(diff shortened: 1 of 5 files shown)


alexandec (approved):

VAMobile/src/utils/jsonFormatting.ts
```diff
@@ -17,3 +17,10 @@ export const fixedWhiteSpaceString = (text: string | undefined) => {
   }
   return ''
 }
+
+export const fixSpecialCharacters = (text: string): string => {
```
alexandec, line 21: A more specific name would be `translateHTMLEntities`. Up to you if you want to change it or not

Sparowhawk: I think we are good with this name, it allows us to fix things that might not be html related in the same function if needed in the future

PR #8278, opened 2024-03-28
bug/8051-Dylan-DemoDirectDepositFix

Sparowhawk:
## Description of Change
Should fix demo mode and e2e tests by manually setting the get bank info query after posting update bank info

## Screenshots/Video
N/A

## Testing
yarn test, e2e tests

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Demo mode and e2e tests should work

## PR Checklist

## For QA


VAMobile/src/api/directDeposit/updateBankInfo.tsx
```diff
@@ -27,10 +27,13 @@ export const useUpdateBankInfo = () => {
 
   return useMutation({
     mutationFn: updateBankInfo,
-    onSuccess: async () => {
+    onSuccess: async (data, newAccountData) => {
       logAnalyticsEvent(Events.vama_update_dir_dep())
       setAnalyticsUserProperty(UserAnalytics.vama_uses_profile())
-      queryClient.invalidateQueries({ queryKey: directDepositKeys.directDeposit })
+      const queryData = queryClient.getQueryData(directDepositKeys.directDeposit) as DirectDepositData
+      newAccountData.financialInstitutionName = 'Bank'
```
alexandec, line 34: Does this mean we're always setting financialInstitutionName to "Bank"? Not sure how this field is normally used, or if it could contain something else that we're overwriting here

VAMobile/src/screens/PaymentsScreen/DirectDepositScreen/DirectDepositScreen.tsx
```diff
@@ -40,9 +41,15 @@ function DirectDepositScreen({ navigation }: DirectDepositScreenProps) {
     isLoading: loading,
     isError: useBankDataError,
   } = useBankData({ enabled: screenContentAllowed('WG_DirectDeposit') && ddNotInDowntime })
-  const bankData = directDepositData?.data.attributes?.paymentAccount
+  const [bankData, setBankData] = useState(directDepositData?.data.attributes?.paymentAccount)
   const { gutter, contentMarginBottom } = theme.dimensions
 
+  useFocusEffect(
```
alexandec, line 47: Is the useFocusEffect needed because the component doesn't re-render when directDepositData changes?

VAMobile/src/api/directDeposit/updateBankInfo.tsx
Sparowhawk: yeah in the original slice we set the data manually to bank when we sent it to the server, so I kept that in line here

VAMobile/src/screens/PaymentsScreen/DirectDepositScreen/DirectDepositScreen.tsx
Sparowhawk: that is correct, when we manually alter the data with setQuery it doesn't force a re render because we navigate backwards. If we left the screen and re entered it would be fine.

alexandec (dismissed):

rbontrager:
I'm pretty sure this change broke direct deposit in staging.  When I save the direct deposit (using the bank info from our detox test) I'm getting the following screen.  Navigating back to the payments screen and back to direct deposit doesn't seem to fix it.

alexandec (approved):

rbontrager (approved):

PR #8284, opened 2024-03-29
feature/8126-Chika-UseUpcomingDaysFromApiInAppointmentsHomeScreenModule

cadibemma:
## Description of Change

- The appointments API sends the number of days alongside the upcoming appointments count. We should update the module to use the number of days coming from the API instead of hard-coding it.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

- The appointments' module is updated to use the `upcomingDaysLimit` field from the API instead of hard-coding it

## PR Checklist

## For QA


dumathane (approved):

VAMobile/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -257,10 +258,13 @@ export function HomeScreen({}: HomeScreenProps) {
             </Box>
           ) : (
             <Box gap={theme.dimensions.condensedMarginBetween} mx={theme.dimensions.condensedMarginBetween}>
-              {!!upcomingAppointmentsCount && (
+              {!!upcomingAppointmentsCount && upcomingDaysLimit && (
```
theodur, line 261:
```suggestion
              {!!upcomingAppointmentsCount && !!upcomingDaysLimit && (
```
This shouldn't ever happen, but if the `upcomingDaysLimit` was returned as `0`, which is also the default value in the Redux store, this would prevent an error from being thrown

theodur (changes requested):

TKDickson:
[Need to update the health screen as well](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8126#issuecomment-2038193966)

TKDickson (approved):
Good to merge to feature branch

PR #8341, opened 2024-04-10
bug/8340-Dylan-FixClaimsAndAppealsMissingAnalytics

Sparowhawk:
## Description of Change
Adds back the claims analytics that was mistakenly removed from the claims and appeals PR

## Screenshots/Video
N/A

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Claims and appeals analytics are functioning like they were pre RQ

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.tsx
+++ b/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.tsx
@@ -48,10 +48,23 @@ function ClaimsHistoryScreen({ navigation }: IClaimsHistoryScreen) {
     data: claimsAndAppealsListPayload,
     isError: claimsAndAppealsListError,
     isLoading: loadingClaimsAndAppealsList,
+    isFetched: finishedLoadingClaimsAndAppeals,
     refetch: refetchClaimsAndAppealsList,
   } = useClaimsAndAppeals(claimType, 1, {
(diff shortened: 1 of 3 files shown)


VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.tsx
```diff
@@ -48,11 +48,24 @@ function ClaimsHistoryScreen({ navigation }: IClaimsHistoryScreen) {
...
     enabled: claimsAndAppealsAccess && (claimsNotInDowntime || appealsNotInDowntime),
   })
 
+  useEffect(() => {
+    if (finishedLoadingClaimsAndAppeals) {
+      const claimsList = claimsAndAppealsListPayload?.data
+      logAnalyticsEvent(
+        Events.vama_claim_count(
+          claimType === ClaimTypeConstants.CLOSED ? claimsList?.length || 0 : 0,
+          claimType === ClaimTypeConstants.ACTIVE ? claimsList?.length || 0 : 0,
```
dumathane, line 63: ...this doesn't feel right? These would log the same numbers?

Sparowhawk: nope, we no longer prefetch both active and closed claims, we only fetch what is needed which is the active ones initially. So to keep this same analytic we can only log one or the other. I do think it would be potentially better to log different analytics for active and closed going forward. Just not sure who to run this by.

dumathane (approved):

PR #8433, opened 2024-04-19
bug/8432-Dylan-DowntimeWindowsRaceCondition

Sparowhawk:
## Description of Change
Looks like there was a race condition on the syncscreen to home screen for downtime windows causing api's to be called when a feature was in downtime. Added the downtime check to the login screen so that we can likely catch this issue sooner until we can convert this to RQ. Also added check to nametag for service history downtime and conditionalized it that if there is no name and no service history that the nametag minimum height is undefined so that it properly hides the nametag.

## Screenshots/Video
N/A

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
HomeScreen should respect downtime windows and a race condition should not exist

## PR Checklist

## For QA


dumathane (approved):

theodur (changes requested):
Tested this locally, and I confirmed it still doesn't work. I manualy set the query to `false` claims query on the HomeScreen just to see, and it still fetches the claimsAndAppeals data, so I think it's an issue with the `enabled` option being `false` not taking effect for whatever reason

Sparowhawk:
> Tested this locally, and I confirmed it still doesn't work. I manually set the query to `false` claims query on the HomeScreen just to see, and it doesn't fetch the data, but the `claimsInDowntime` const is already set to `true` by the time it hits the query on the home screen, so not sure why the logic is still not working
> 
> Edit: So the logic on the home screen is working correctly, but when you navigate to the feature, it's still showing claims data rather than showing the downtime message. I think one of the screens using the claims query, isn't disabling it when the downtime is true.
> 
> Edit 2: I think a more bulletproof way of preventing queries from running when the feature is in downtime or a user isn't authorized, is to incorporate that logic directly into the custom hook, so that we don't have to manually add that every place a query is used. Because it's easy to miss adding that when using a custom hook
> 
> Edit 3: I also don't think we need to check for downtime on both the sync and login screen, since that creates duplicate calls. Ideally, it would be more efficient to keep the downtime check on the sync screen instead of the login screen, because with the `isFocused` logic on the login screen, every time that screen is shown, it'll call the downtime endpoint, which will naturally increase how many times that API is called. If it's a matter of a race condition, we can add a loading state to the downtime check and make that a requirement before completing sync

Rachael was also testing this for me and it was working on how this PR is currently setup, it does seem finicky for sure.

I do see that you opened your own PR for this issue. Going to leave a comment there to include my nametag changes for the min height in the event the query for military info doesn't run and we don't have a full name for whatever  reason. Otherwise it will just be a blue bar that goes across the screen.

I agree this makes more sense to run both the authorized and downtime checks on the api calls itself so that we don't have to remember to add them everywhere the api is called. The only caveat is that for both of those there are still error states that we need to display and just wanted to call that out.

I agree with your third edit and in a perfect world I would just complete the migration for that API to RQ so we automatically get those flags. The whole reason of adding it to the Login screen was to purposely make as little changes as possible and one extra api call isn't the end of the world until we can get the migration to RQ to handle all the actual work on the sync screen in order to block progressing to the home screen until the api is complete.

PR #8451, opened 2024-04-23
bug/8442-Dylan-EUAnalyticFix

Sparowhawk:
## Description of Change
Updated EU analytics to be in a useEffect instead of on render to reduce the amount of analytics being reported as a result of rerendering from RQ and HSP

## Screenshots/Video
Screenshot 2024-04-23 at 10 52 50 AM

See only 1 vama_eu_shown analytic in that log

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
only one analytic call for vama_eu_shown should be called per home screen view on startup

## PR Checklist

## For QA


VAMobile/src/components/EncourageUpdate.tsx
```diff
@@ -61,6 +61,19 @@ export const EncourageUpdateAlert = () => {
...
 
+  useEffect(() => {
+    if (
+      featureEnabled('inAppUpdates') &&
+      storeVersion &&
+      localVersionName &&
+      skippedVersion &&
+      skippedVersion !== storeVersion &&
+      ((isIOS() && storeVersion > localVersionName) || (!isIOS() && +storeVersion > +localVersionName))
+    ) {
```
theodur, line 72: It'd be nice to have this logic stored in a variable since it's used in multiple places and is a pretty beefy conditional

Sparowhawk: updated

theodur (dismissed):

rbontrager (approved):

rbontrager (approved):

PR #8481, opened 2024-04-26
bug/8432-theo-ReactQueryMaintenanceWindows

theodur:
## Description of Change

Updates custom React Query hooks to directly include the downtime maintenance and authorized services logic.

Note: I only updated relevant screens that would be experiencing API calls being made when they're not supposed to be. 

## Screenshots/Video

N/A

## Testing

Tested locally on staging & demo mode

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

Verify API for features migrated to React Query (Rx, Claims & Appeals, etc) aren't being called when the feature is in downtime.

## PR Checklist

## For QA


VAMobile/src/api/appointments/getAppointments.tsx
```diff
@@ -36,8 +39,13 @@ export const useAppointments = (
   page: number,
   options?: { enabled?: boolean },
 ) => {
+  const { data: authorizedServices } = useAuthorizedServices()
+  const appointmentsInDowntime = useDowntime('appointments')
+  const queryEnabled = options && has(options, 'enabled') ? options.enabled : true
```
theodur, line 44: This conditional looks funky, but we need to check whether the `options` object has the `enabled` property instead of just checking `options?.enabled`. Mainly because when the `enabled` option isn't set on the screen that's using this hook, its value will result in `undefined`, which would make this conditional `false`, but we want it to run by default if the `enabled` option isn't specified (assuming the feature isn't in maintenance and the user have access to the feature).

I was also considering turning this into a helper function that all queries can reference, but I manually added it to the queries for now since the logic is fairly simple, and we're only using this in a handful of queries.

VAMobile/src/api/appointments/getAppointments.tsx
```diff
@@ -36,8 +39,13 @@ export const useAppointments = (
   page: number,
   options?: { enabled?: boolean },
 ) => {
+  const { data: authorizedServices } = useAuthorizedServices()
+  const appointmentsInDowntime = useDowntime('appointments')
```
theodur, line 43: Unfortunately, using `DowntimeFeatureTypeConstants` here caused unit tests to fail, because they don't recognize the constant. This was also [pointed out](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/6634fd15d77afdb001620839e67e1acdb784711e/VAMobile/src/screens/SyncScreen/SyncScreen.tsx#L40) a while ago. We'll need to look into this

VAMobile/src/api/authorizedServices/getAuthorizedServices.tsx
```diff
@@ -24,5 +24,6 @@ export const useAuthorizedServices = (options?: { enabled?: boolean }) => {
     meta: {
       errorName: 'getAuthorizedServices: Service error',
     },
+    staleTime: 300000, // 5 minutes
```
theodur, line 27: Since we'll be calling this query in many other queries, I bumped up the stale time to 5 minutes. There's an argument to be made for increasing it even more, but we can adjust this as we see fit

VAMobile/src/testUtils.tsx
```diff
@@ -189,6 +190,24 @@ function render(ui, { preloadedState, navigationProvided = false, queriesData, .
...
+      genderIdentity: true,
+      lettersAndDocuments: true,
+      militaryServiceHistory: true,
+      paymentHistory: true,
+      preferredName: true,
+      prescriptions: true,
+      scheduleAppointments: true,
+      secureMessaging: true,
+      userProfileUpdate: true,
+    })
```
theodur, line 210: Since we're referencing authorized services in multiple queries, this was needed to prevent multiple unit tests from failing because the values for these weren't mocked. These values can still be altered if need be by using setting the `queriesData` param just like we do with other data

Sparowhawk (dismissed):
Just calling out from my own PR for this. The nametag component min height needs to be conditionalized as there is a scenario where we do not end up with a full name and the military info api is down. This results in just a solid blue banner across the screen and is not desired, so we should just hide it in this scenario.

dumathane (dismissed):

dumathane (approved):

Sparowhawk (approved):

DJUltraTom (approved):

PR #8506, opened 2024-04-30
finished

Sparowhawk:
## Screenshots/Video
Before: 
image

After:

https://github.com/department-of-veterans-affairs/va-mobile-app/assets/87150991/fbd0d525-d42e-4572-8433-c394e9872c83

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Remove inline links and add them after the message using the new link component

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/utils/secureMessaging.tsx
+++ b/VAMobile/src/utils/secureMessaging.tsx
@@ -1,9 +1,11 @@
 import React, { ReactNode } from 'react'
 import DocumentPicker from 'react-native-document-picker'
-import { TouchableWithoutFeedback } from 'react-native-gesture-handler'
 import { Asset, launchCamera, launchImageLibrary } from 'react-native-image-picker'
 import { ImagePickerResponse } from 'react-native-image-picker/src/types'
 
+import { Link } from '@department-of-veterans-affairs/mobile-component-library'
+import { IconProps } from '@department-of-veterans-affairs/mobile-component-library/src/components/Icon/Icon'
+import { LinkProps } from '@department-of-veterans-affairs/mobile-component-library/src/components/Link/Link'
 import { ActionSheetOptions } from '@expo/react-native-action-sheet'
 import { TFunction } from 'i18next'
 import _ from 'underscore'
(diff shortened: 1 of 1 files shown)


VAMobile/src/utils/secureMessaging.tsx
```diff
@@ -510,22 +505,19 @@ export const getLinkifiedText = (
...
+          text: previousText + ' ' + text + ' ' + nextText,
+          type: 'call',
+          phoneNumber: previousText + text + nextText,
+          a11yLabel: getNumberAccessibilityLabelFromString(previousText + text + nextText),
+          a11yHint: t('openInPhoneMessaging.a11yHint'),
+        }
+
+        linkCollection.push(
+          <Box mb={theme.dimensions.standardMarginBetween}>
+            <Link icon={{ preventScaling: true }} {...linkProps} />
```
alexandec, line 518: Please use LinkWithAnalytics instead of Link for all of these. That way we get analytics for free (props are the same)

VAMobile/src/utils/secureMessaging.tsx
```diff
@@ -537,90 +529,95 @@ export const getLinkifiedText = (
...
-          accessibilityLabel={text}
-          accessibilityHint={t('openInEmailMessaging.a11yHint')}>
-          <TextView variant="MobileBodyLink">{text}</TextView>
-        </TouchableWithoutFeedback>,
+      const iconProps: IconProps = {
+        name: 'ExternalLink',
+      }
+      const linkProps: LinkProps = {
+        text: text,
+        type: 'custom',
```
alexandec, line 537: Can't this use the 'url' type since it's a URL?

VAMobile/src/utils/secureMessaging.tsx
```diff
@@ -537,90 +529,95 @@ export const getLinkifiedText = (
...
```
alexandec, line 557: Can't this use the 'url' type since it's a URL?

VAMobile/src/utils/secureMessaging.tsx
Sparowhawk: updated

VAMobile/src/utils/secureMessaging.tsx
Sparowhawk: good call, updated

VAMobile/src/utils/secureMessaging.tsx
Sparowhawk: good call, updated

alexandec (approved):

PR #8523, opened 2024-05-02
Automation/6261-Chanel-Require-Automate-QA-Approval-Before-PR-Merge

IsraelleHub:
## Description of Change
This  should be a Github actions workflow that:

Runs on each PR Open
Runs on each PR "approval"
Only runs when merges are going into develop
Only runs on code that requires both FE and QA review
Also:

It should run immediately with a failure, blocking merged for code QA needs to review
On each approval, it should rerun to test that there are at least 3 approvals, one of which is from a member of the QA team ([flagship-mobile-qa](https://github.com/orgs/department-of-veterans-affairs/teams/flagship-mobile-qa/members)). Ideally, if we can check team membership that would be best, but if not having users listed directly is fine.
the [CODEOWNERS](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/.github/CODEOWNERS) file outlines areas where the code needs QA approval for merge
Include code that touches the app:

VAMobile/src
VAMobile/e2e
VAMobile/package.json
VAMobile/.detoxrc.json

Exclude PRs with only unit tests:
VAMobile/src/*.test.tsx

## Screenshots/Video
No UI changes.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/.github/workflows/pr_checks_for_qa_approvals.yml
+++ b/.github/workflows/pr_checks_for_qa_approvals.yml
@@ -15,9 +15,9 @@ on:
   pull_request_review:
     types: [submitted, dismissed]
 
-  pull_request_target:
-    branches:
-      - 'develop'
+  #pull_request_target:
+    #branches:
+      #- 'develop'
   #push: 
    # branches: 
    #   - 'Automation/6261-Chanel-Require-QA-Approval'    
--- a/.github/workflows/pr_checks_for_qa_approvals.yml
+++ b/.github/workflows/pr_checks_for_qa_approvals.yml
@@ -18,6 +18,9 @@ on:
   pull_request_target:
     branches:
       - 'develop'
+  push: 
+    branches: 
+      - 'Automation/6261-Chanel-Reuire-QA-Approval'    
 
 jobs:
   check_for_qa_approval:
--- a/.github/workflows/pr_checks_for_qa_approvals.yml
+++ b/.github/workflows/pr_checks_for_qa_approvals.yml
@@ -1,0 +1,67 @@
+
+#
+# Checks if QA approvals are necessary, and if so runs the associated action when a new review is added
+#
+
+name: '[Utils] QA required approval'
+
+on:
+  pull_request:
+    types: [opened, edited]
+  pull_request_review:
+    types: [submitted, dismissed]
+  push:
+
+jobs:
+  check_for_qa_required:
+    if: github.event.review.state == 'APPROVED'
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_checks_for_qa_approvals.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          token=${{ secrets.GITHUB_TOKEN }}
+          base64AuthInfo='$token | base64 -e'
+          response=$(curl --request GET \
+          --url https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/reviews \
+          --header 'Authorization: $base64AuthInfo' \
+          --header 'Content-Type: application/json')
+          approvals = $response | jq '.[] | select(.state=="APPROVED")'
+          echo $approvals
+          if $aprovals.length > 2
```
Sparowhawk, line 49: just wanted to let you know that we just currently require 1 engineer approval

.github/workflows/pr_checks_for_qa_approvals.yml
```diff
@@ -0,0 +1,67 @@
...
+                echo ::set-output name=has_qa_approval::'true'
+              else 
+                echo 'This PR requires QA approval to merge'
+                echo ::set-output name=has_qa_approval::'false'
+              fi
+          else
+            echo 'This PR requires 3 approvals, including one QA approval, before merging.'
+            echo ::set-output name=has_qa_approval::'false'
+          fi
+  add_or_remove_qa_required_label:
```
Sparowhawk, line 62: we already have an auto labeling workflow that checks if QA approved or not and applies a ready to merge label, otherwise with just engineer approval it applies FE-with QA so you don't need to do any labeling on the pr, you can leave it so that this workflow only does the blocking of merging

IsraelleHub: Awesome. Thank you for the feedback @Sparowhawk

.github/workflows/pr_checks_for_qa_approvals.yml
IsraelleHub: Great. I will update the code.

IsraelleHub: @Sparowhawk does that mean it now require one approval from qa and one approval from engineer>

Sparowhawk: that is correct. It only requires 1 approval from QA and one other approval(engineer for code related changes that users would actually see, anyone for documentation or other items that users wouldn't see)

PR #8541, opened 2024-05-03
bug/Binny-8537-Three-BitterHeadingPixelChange

dumathane:
## Screenshots/Video

Before/After
&nbsp;&nbsp;

Places changed:
image

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/styles/themes/standardTheme.ts
+++ b/VAMobile/src/styles/themes/standardTheme.ts
@@ -64,6 +64,10 @@ const fontSizes = {
     fontSize: 22,
     lineHeight: 24,
   },
+  BitterHeading26: {
+    fontSize: 26,
+    lineHeight: 26,
+  },
   BitterBoldHeading: {
     fontSize: 26,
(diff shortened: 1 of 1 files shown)


VAMobile/src/styles/themes/standardTheme.ts
```diff
@@ -64,6 +64,10 @@ const fontSizes = {
     fontSize: 22,
     lineHeight: 24,
   },
+  BitterHeading26: {
+    fontSize: 26,
+    lineHeight: 26,
+  },
```
theodur, line 70: Instead of making this new font size, I think we should change the values in `BitterHeading`, and make a new font size for large nav/activity buttons since they're the only things using the values from `BitterHeading`. That font size for the buttons could be called `ButtonHeader` or `LargeButtonHeader`, etc.

The `CategoryBitterHeading` variant can also be removed, it's not being used anywhere.

PR #8599, opened 2024-05-10
bug/8578-Three-DeepLinkOptimizations

theodur:
## Description of Change

There is a delay whenever an activity module is pressed on the home screen. This is due to a delay in the deep link setting the screen stack state for the screen it's navigating to (essentially renders all screens leading up to the destination screen as if a user manually navigated to it). Since this isn't caused by an API call, there isn't anything we could tie a loading component to. Instead, I tried to make some rendering optimizations to the screens that get rendered in tangent to the destination screen. This probably won't help much, but it's at least some type of efficiency improvement.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/api/facilities/getFacilitiesInfo.tsx
+++ b/VAMobile/src/api/facilities/getFacilitiesInfo.tsx
@@ -24,5 +24,6 @@ export const useFacilitiesInfo = () => {
     meta: {
       errorName: 'getFacilitiesInfo: Service error',
     },
+    staleTime: Infinity,
   })
 }
--- a/VAMobile/src/screens/HealthScreen/HealthScreen.tsx
+++ b/VAMobile/src/screens/HealthScreen/HealthScreen.tsx
@@ -47,6 +47,7 @@ export function HealthScreen({}: HealthScreenProps) {
   const theme = useTheme()
   const navigateTo = useRouteNavigation()
   const { t } = useTranslation(NAMESPACE.COMMON)
+  const isFocused = useIsFocused()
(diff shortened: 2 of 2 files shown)


VAMobile/src/api/facilities/getFacilitiesInfo.tsx
```diff
@@ -24,5 +24,6 @@ export const useFacilitiesInfo = () => {
     meta: {
       errorName: 'getFacilitiesInfo: Service error',
     },
+    staleTime: Infinity,
```
theodur, line 27: This was being called every time the Health screen was opened. Don't think this is likely to change in a user's session, so I set it to never be stale

VAMobile/src/screens/HealthScreen/HealthScreen.tsx
```diff
@@ -108,21 +108,6 @@ export function HealthScreen({}: HealthScreenProps) {
...
-      if (isScreenContentAllowed && cernerExist && ((firstTimeLogin && mixedCerner) || (newSession && allCerner))) {
-        navigateTo('HealthHelp')
-        await AsyncStorage.setItem(FIRST_TIME_LOGIN, '')
-        await AsyncStorage.setItem(NEW_SESSION, '')
-      }
-    }
-
-    healthHelpScreenCheck()
-  }, [allCerner, cernerExist, isScreenContentAllowed, mixedCerner, navigateTo])
-
```
theodur, line 125: This duplicate effect was probably accidentally added in a merge

dumathane (approved):

PR #8613, opened 2024-05-10
bug/8465-Dylan-RQNotAuthorizedScreensPrioritization

Sparowhawk:
## Description of Change
Changed the priority on how sub screens are rendered. New priority on every feature is as follows
1. Downtime
2. Loading
3. user authorized services error with retry
4. user not authorized screen
5. other api errors with retry(terms and conditions is higher priority over other api errors for SM)
6. no data screens
7. actual screen assuming everything goes well

## Screenshots/Video
N/A

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Missed this originally because of how I was mocking my data. With RQ it looks the feature's api is called even when the authorizedServices for that feature is set to false. This is causing the default error screen to appear instead of the various you don't have permission for this specific feature screen.

## PR Checklist

## For QA


VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.tsx
```diff
@@ -122,18 +122,24 @@ function ClaimsHistoryScreen({ navigation }: IClaimsHistoryScreen) {
...
         <LoadingComponent text={t('claimsAndAppeals.loadingClaimsAndAppeals')} />
-      ) : claimsAndAppealsListError ||
-        getUserAuthorizedServicesError ||
-        (!claimsNotInDowntime && !appealsNotInDowntime) ? (
+      ) : getUserAuthorizedServicesError ? (
         <ErrorComponent
           onTryAgain={fetchInfoAgain}
           screenID={ScreenIDTypesConstants.CLAIMS_HISTORY_SCREEN_ID}
-          error={claimsAndAppealsListError}
+          error={getUserAuthorizedServicesError}
```
theodur, line 133: We shouldn't be displaying the error component when the authorized services endpoint has an error, because if that endpoint were to fail, we don't want to block a user from using features in the app. We (Global) had a similar discussion for some HSP work, and decided if that endpoint fails, we should still allow the user to attempt to access features

VAMobile/src/screens/BenefitsScreen/Letters/LettersListScreen.tsx
```diff
@@ -127,24 +128,31 @@ function LettersListScreen({ navigation }: LettersListScreenProps) {
...
+        <ErrorComponent screenID={ScreenIDTypesConstants.LETTERS_LIST_SCREEN_ID} />
+      ) : loading || loadingUserAuthorizedServices ? (
         <LoadingComponent text={t('letters.list.loading')} />
-      ) : errorCheck || !lettersNotInDowntime ? (
+      ) : getUserAuthorizedServicesError ? (
+        <ErrorComponent
+          screenID={ScreenIDTypesConstants.LETTERS_LIST_SCREEN_ID}
+          error={getUserAuthorizedServicesError}
+          onTryAgain={refetchAuthServices}
+        />
```
theodur, line 146: Same here and the other files that were updated in this PR. We shouldn't be displaying the error component when the authorized services endpoint has failed

theodur (changes requested):

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.tsx
```diff
@@ -122,18 +122,24 @@ function ClaimsHistoryScreen({ navigation }: IClaimsHistoryScreen) {
       backLabelOnPress={navigation.goBack}
       title={title}
       testID="claimsHistoryID">
-      {loadingClaimsAndAppealsList || loadingUserAuthorizedServices ? (
+      {!claimsNotInDowntime && !appealsNotInDowntime ? (
+        <ErrorComponent screenID={ScreenIDTypesConstants.CLAIMS_HISTORY_SCREEN_ID} />
+      ) : loadingClaimsAndAppealsList || loadingUserAuthorizedServices ? (
```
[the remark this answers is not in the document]
Sparowhawk: Unfortunately it will be needed. In theory you are correct, but if this api fails on the sync screen and it gets to the following screens it would be needed, and primarily if there is an error that is being refreshed. Otherwise what happens is they hit the refresh button and then there is no loading screen before either an error returns or the screen fixes itself randomly. So the loading screen is needed in those instances.

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimsHistoryScreen/ClaimsHistoryScreen.tsx
Sparowhawk: interesting, I will make those changes then

Sparowhawk: I guess my issue now is that if that endpoint fails, then they would technically be displayed the not authorized screen due to !userAuthorizedServices?.militaryServiceHistory ?

VAMobile/src/screens/BenefitsScreen/Letters/LettersListScreen.tsx
Sparowhawk: same as my comment above, I disagree with this statement and we may need product to weigh in. Cause what will happen is the not authorized screen will then display instead causing a worse experience since that does not have a refresh

Sparowhawk:
started a slack thread to get products decision for your last 2 comments

theodur (approved):

PR #8621, opened 2024-05-13
feature/Binny-8620-VetStatusAnalytics

dumathane:
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/components/Nametag/Nametag.tsx
+++ b/VAMobile/src/components/Nametag/Nametag.tsx
@@ -7,9 +7,9 @@ import { useServiceHistory } from 'api/militaryService'
 import { usePersonalInformation } from 'api/personalInformation/getPersonalInformation'
 import { BranchesOfServiceConstants, ServiceHistoryData } from 'api/types'
 import { BackgroundVariant, Box, TextView, VAIcon } from 'components'
-import { UserAnalytics } from 'constants/analytics'
+import { Events, UserAnalytics } from 'constants/analytics'
 import { NAMESPACE } from 'constants/namespaces'
-import { setAnalyticsUserProperty } from 'utils/analytics'
+import { logAnalyticsEvent, setAnalyticsUserProperty } from 'utils/analytics'
 import { useRouteNavigation, useTheme } from 'utils/hooks'
 
 export const Nametag: FC = () => {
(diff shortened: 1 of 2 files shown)


VAMobile/src/components/Nametag/Nametag.tsx
```diff
@@ -39,6 +39,11 @@ export const Nametag: FC = () => {
   serviceHistory.forEach((service) => {
     if (service.honorableServiceIndicator === 'Y') {
       showVeteranStatus = true
+      logAnalyticsEvent(Events.vama_vet_status_shown())
+    } else if (service.honorableServiceIndicator === 'N') {
+      logAnalyticsEvent(Events.vama_vet_status_nStatus())
+    } else if (service.honorableServiceIndicator === 'Z') {
+      logAnalyticsEvent(Events.vama_vet_status_zStatus(service.characterOfDischarge))
```
theodur, line 46: To prevent this from logging multiple times when re-renders happen, this should be in its own `useEffect`. Even then, it could potentially still be logged twice, since the Nametag component gets rendered on the Home and Profile screens. 

To truly only log this once, you'd either need a global place to track whether this has been logged before, and add that result as a condition in these `if` statements, or to move logging this to a different location where it wouldn't get re-rendered, like the Sync screen since we prefetch the service history there. There's also the option of logging this directly in the query function, especially if we were to set the stale time of this query to `Infinity`, because realistically a user's service history isn't likely to change during a session, which means that query function would only be ran once.

theodur: Also, not sure if this would make it easier to track, but instead of having separate events for logging the different service indicators, we could have one event called `vama_vet_status_indicator` where the indicator is a required parameter, and the character of discharge as an optional parameter

timwright12 (dismissed):
test approval

cadibemma (approved):
Looks good!

dumathane:
I checked Theo's edit to my branch (Thank you theo-wan Kenobi!) and it all works as expected now. Logging in with demo mode twice results in 1 log of analytics, and then with judy twice also just one log of analytics. No noticeable differences in slowdowns.

theodur (approved):

TKDickson (approved):

PR #8635, opened 2024-05-14
bug/8584-theo-RQOptimizations

theodur:
## Description of Change

Makes some React Query optimizations. Changes include:
- Disabling query retries
- Busting the backend cache for appointments, claims & appeals, secure messages, and vaccines to ensure we're always retrieving fresh data whenever we fetch from those APIs
- Managing pagination in the app for the above APIs

During my testing, I noticed for the vaccines and claims lists, the screens weren't auto-scrolling to the top when cycling through pages in the pagination, so I added that functionality for those screens.

## Screenshots/Video

N/A

## Testing

- Tested appointments and vaccines pagination in staging with Judy
- Tested claims and appeals pagination in staging with Arfan
- Tested secure messages in staging with user 226

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

- Drop react query retries to zero. (arguably, not related to caching)
- Drop BE caching all together and cache on frontend (to avoid loading screens for no reason)
- Ensure pagination is handled correctly

## PR Checklist

## For QA


dumathane (approved):

Sparowhawk (commented):
So far my notes with playing around
If HSP is loading on the home screen the profile button doesn’t work, it does work once everything is loaded in
Having more than 20 claims and appeals starts making the api take significantly longer. I tested on +54 and was able to navigate to claims and still had to wait about 4-5 seconds even though it was called on the home screen to start
Past claims and appeals there are 250 for +54 and it took about 18 seconds to load, we might want to prefetch these in the background, but not actually use them for anything on the home screen to reduce this load time.
Currently looking at appts to see how bad that loading is





Dylan Nienberg
  10 minutes ago
4. Appointments past loading is around 30 seconds just for the past 3 months which has 87 appts
5. Similar issues for all the other past appt ranges
We likely should just prefetch all those on the home screen as well for each timeframe to reduce load time for users


Dylan Nienberg
  [9 minutes ago](https://adhoc.slack.com/archives/C063KLSMVAR/p1715874784814479?thread_ts=1715866896.649389&cid=C063KLSMVAR)
6. upcoming appts does take forever

Sparowhawk:
Another thought is for past appt timeframes and past appts we could increase the stale time significantly as that data should not change until the next day

Sparowhawk:
Code does look good though if we can fix these loading issues

theodur:
Based on our conversations in Slack and DSU, I made some updates to prefetch past appointments (3 months) and closed claims whenever we're fetching upcoming appointments and active claims. This also won't impact the loading/error state on the home screen, meaning if the prefetch call for past appointments or closed claims fails, it won't show an error on the home screen, since all we care about for the home screen UI is upcoming appointments and active claims

Sparowhawk (approved):

Sparowhawk (approved):

TKDickson (approved):
approved to merge to feature branch

PR #8636, opened 2024-05-14
updated

Sparowhawk:
## Description of Change
Updated the regex to be more inclusive of extensions and to stop if punctuation isn't followed correctly. Defaulting to not have a link

## Screenshots/Video
Screenshot 2024-05-14 at 4 15 51 PM

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Shouldn't have broken phone links in the app that wrap poorly due to typos.

## PR Checklist

## For QA


alexandec:
Recommend adding a unit test or two to check for this new case

Sparowhawk:
added unit test

alexandec (dismissed):

Sparowhawk:
Ok

I fixed other whitespace character like tabs or enters not linking text

I also fixed all of them aside from this format

(800) 555-1234

Sparowhawk:
Checking collections of 2 strings ended up being pretty bad so we have to exclude that scenario at least for now

alexandec (approved):

rbontrager (approved):

PR #8679, opened 2024-05-20
feature/8582-theo-AnalyticsForRQRetries

theodur:
## Description of Change

Adds a new analytic event `vama_react_query_retry`, for logging how often query retries are successful. I am using the contact information query for this experiment. This event will only be logged if the contact information query has failed, and a retry occurred (which should always happen on a failure since I set the retry option to `1` for that query). It will log whether the retry was `successful` or a `fail`.

## Screenshots/Video

N/A

## Testing

Tested locally by updating the client API function to fail on the first fetch.

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/ContactInformationScreen.tsx
+++ b/VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/ContactInformationScreen.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Pressable } from 'react-native'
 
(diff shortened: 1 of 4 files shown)


VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/ContactInformationScreen.tsx
```diff
@@ -132,9 +132,22 @@ function ContactInformationScreen({ navigation }: ContactInformationScreenProps)
...
+
+  useEffect(() => {
+    if (failureCount > 0) {
+      setRetried(true)
+    }
+
+    if (retried && !loadingContactInformation) {
+      const retryStatus = contactInformationError ? 'fail' : 'success'
+      logAnalyticsEvent(Events.vama_react_query_retry(retryStatus))
+    }
```
dumathane, line 149: Is there a reason we couldn't just do if `(failureCount > 0 && !loadingContactInformation)` ?

dumathane:
What steps did you take to test this, Theo? Can you detail them.

dumathane:
Tested locally by updating the client API function to fail on the first fetch.

I see this, just interested what this looked like you for you specifically in testing.

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/ContactInformationScreen.tsx
theodur: Good question. That's what I wanted to do initially, but after a retry succeeds, the failure count gets wiped, so it wouldn't tell is if the query was retried when logging successful retries

theodur:
> Tested locally by updating the client API function to fail on the first fetch.
> 
> I see this, just interested what this looked like you for you specifically in testing.

@dumathane I modified the `get` function to fail the first time it receives the contact info endpoint

```ts

let succeed = false

export const get = async function <T>(
  endpoint: string,
  params: Params = {},
  abortSignal?: AbortSignal,
): Promise<T | undefined> {
  if (endpoint === '/v0/user/contact-info' && !succeed) {
    succeed = true
    throw new Error('fail')
  }
  return call<T>('GET', endpoint, params, undefined, abortSignal)
}

dumathane (approved):

TKDickson:
Get Contact Info is called in letters, too -- on the screen that shows the mailing address where the letter is being sent. Can you make sure that we log the retries analytic there too?

TKDickson (approved):
approved to merge to develop

PR #8682, opened 2024-05-21
add json api spec BE docs 

aherzberg:
## Description of Change
add json api spec BE docs 

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/documentation/docs/Engineering/BackEnd/Architecture/JSONAPI.md
+++ b/VAMobile/documentation/docs/Engineering/BackEnd/Architecture/JSONAPI.md
@@ -1,0 +1,11 @@
+# JSON API 
+
+## Overview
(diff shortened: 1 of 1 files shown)


VAMobile/documentation/docs/Engineering/BackEnd/Architecture/JSONAPI.md
aherzberg, line 1: The ticket calls for specifying which part of the spec we care about but we seem to be following the whole spec except for the exceptions I listed so It's probably worth trying to adhere to the whole spec. I also initially listed more information like the version of the spec we adhere to but that seemed like overkill. They say that all future versions will be backwards compatible so we should just always adhere to the latest version.

Tonksthebear (approved):
LGTM

PR #8694, opened 2024-05-22
8130-rachael-detox-add-platform-name-to-default-test-run

rbontrager:
## Description of Change

Add the ability to add the platform name to default test run. This was done by adding an option to the run workflow.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/.github/workflows/update_testrail_results.yml
+++ b/.github/workflows/update_testrail_results.yml
@@ -16,6 +16,12 @@ on:
       testRail_name:
         type: string
         default: ''
+      test_specific_OS_needed:
+        type: boolean
+        default: false
+      test_OS_name:
+        type: string
+        default: ''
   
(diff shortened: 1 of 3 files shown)


.github/workflows/update_testrail_results.yml
```diff
@@ -16,6 +16,12 @@ on:
       testRail_name:
         type: string
         default: ''
+      test_specific_OS_needed:
+        type: boolean
+        default: false
+      test_OS_name:
+        type: string
+        default: ''
```
theodur, line 24: I don't think you need two inputs for this. You should be able to just use `test_OS_name` and reference that directly since it already has a default value

theodur (approved):

PR #8747, opened 2024-05-30
bug/8746-theo-LoadingAnimationGlitches

theodur:
## Description of Change

Prevents the loading icon from re-rendering whenever the home screen re-renders. I also refactored the home screen to make the component easier to read, which is what most of the changes in this PR are.

### Before
[LoadingAnimationGlitching.webm](https://github.com/department-of-veterans-affairs/va-mobile-app/assets/9042882/8eab304e-f00c-4bec-9956-4f8f14791bec)

### After
Note: The error is unrelated to these changes, one of the other APIs just happened to have an issue at the time.

[LoadingAnimationGlitchingAfter.webm](https://github.com/department-of-veterans-affairs/va-mobile-app/assets/9042882/33c3ef2d-3423-423f-ab62-65ae3ffdf32e)

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HomeScreen/HomeScreen.tsx
+++ b/VAMobile/src/screens/HomeScreen/HomeScreen.tsx
@@ -51,7 +51,7 @@ import { a11yLabelVA } from 'utils/a11yLabel'
 import { logAnalyticsEvent } from 'utils/analytics'
 import getEnv from 'utils/env'
 import { roundToHundredthsPlace } from 'utils/formattingUtils'
-import { useAppDispatch, useDowntime, useRouteNavigation, useTheme } from 'utils/hooks'
+import { useDowntime, useRouteNavigation, useTheme } from 'utils/hooks'
 
 import ContactVAScreen from './ContactVAScreen/ContactVAScreen'
 import { HomeStackParamList } from './HomeStackScreens'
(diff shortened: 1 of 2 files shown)


VAMobile/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -70,111 +70,124 @@ import NotificationsSettingsScreen from './ProfileScreen/SettingsScreen/Notifica
 const { WEBVIEW_URL_FACILITY_LOCATOR, LINK_URL_ABOUT_PACT_ACT } = getEnv()
 
 type HomeScreenProps = StackScreenProps<HomeStackParamList, 'Home'>
+const MemoizedLoadingComponent = React.memo(LoadingComponent)
```
theodur, line 73: React Query re-renders the parent component whenever the state of a query is updated. Even though the re-renders are limited to the query fields that are referenced in the component, we're referencing quite a bit of fields from multiple queries, which causes multiple re-renders of the parent component, which also causes many child components like `LoadingComponent` to re-render as well. Whenever the `LoadingComponent` is re-rendered, the `Lottie` animation to reset its position, making the component look like it's glitching.

My initial approach was to try and use the `notifyOnChangeProps` option in the queries to limit re-renders. This option prevents the component from re-rendering until any of the listed properties change, e.g. `notifyOnChangeProps['isFetched']` will only re-render the component when the `isFetched` property is changed for that query. This wouldn't work for the Home screen's use case, because even if the Home screen only re-renders when the `isFetched` fields are updated, it would still re-render the animation after each query is finished fetching. 

The best way I could come up with for preventing re-renders was using `React.memo` to prevent the `LoadingComponent` from re-rendering if any of its props haven't changed.

dumathane (approved):

TKDickson (approved):
Approved to merge to feature branch

PR #8824, opened 2024-06-14
bug/8598-Dylan-AndroidTextLeftAlignmentNativeActionSheets

Sparowhawk:
## Description of Change
Change the alignment for Android action sheets only to left. iOS remains unchanged

## Screenshots/Video
Screenshot 2024-06-14 at 12 29 01 PM

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
if this change will result in changing [native iOS action sheets](https://developer.apple.com/design/human-interface-guidelines/action-sheets#Best-practices), then no, let's not do it. We don't want to override the iOS appearance at all.

If you can achieve the fix only for Android, then sure – align left.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/utils/hooks/index.tsx
+++ b/VAMobile/src/utils/hooks/index.tsx
@@ -306,9 +306,17 @@ export function useDestructiveActionSheet(): (props: useDestructiveActionSheetPr
     showActionSheetWithOptions(
       {
         title: props.title,
-        titleTextStyle: { fontWeight: 'bold', textAlign: 'center', color: currentTheme.colors.text.primary },
+        titleTextStyle: {
+          fontWeight: 'bold',
+          textAlign: isIOS() ? 'center' : 'left',
+          color: currentTheme.colors.text.primary,
+        },
         message: props.message,
-        messageTextStyle: { fontWeight: 'normal', textAlign: 'center', color: currentTheme.colors.text.primary },
+        messageTextStyle: {
+          fontWeight: 'normal',
+          textAlign: isIOS() ? 'center' : 'left',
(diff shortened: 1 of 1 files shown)


VAMobile/src/utils/hooks/index.tsx
```diff
@@ -306,9 +306,17 @@ export function useDestructiveActionSheet(): (props: useDestructiveActionSheetPr
     showActionSheetWithOptions(
       {
         title: props.title,
-        titleTextStyle: { fontWeight: 'bold', textAlign: 'center', color: currentTheme.colors.text.primary },
+        titleTextStyle: {
+          fontWeight: 'bold',
+          textAlign: isIOS() ? 'center' : 'left',
```
alexandec, line 311: Instead of performing the check multiple times, can we have a function or a constant to ensure consistency?

VAMobile/src/utils/hooks/index.tsx
```diff
@@ -46,6 +46,7 @@ import { capitalizeFirstLetter, stringToTitleCase } from 'utils/formattingUtils'
 import { isAndroid, isIOS, isIpad } from 'utils/platform'
 import { WaygateToggleType, waygateNativeAlert } from 'utils/waygateConfig'
 
+const textAlightment = isIOS() ? 'center' : 'left'
```
alexandec, line 49: Typo "Alightment". Could probably just call it "textAlign" and use shorthand for the object property

alexandec (approved):

rbontrager (approved):

PR #8849, opened 2024-06-17
8603-Chika-AFStyleAPIErrorHandlingCustomization

cadibemma:
## Description of Change

- We need to intercept BE status 418 codes and display a custom error response to users given the following documented data properties: [link](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/Engineering/BackEnd/Features/CustomErrorResponses).

#### BE Response:
image

#### Before / After: 
&nbsp;&nbsp;

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/utils/errors.ts
```diff
@@ -44,6 +44,10 @@ export const getCommonErrorFromAPIError = (error: APIError, screenID?: ScreenIDT
     return CommonErrorTypesConstants.APP_LEVEL_ERROR_DISABILITY_RATING
   } else if (screenID === ScreenIDTypesConstants.APPOINTMENTS_SCREEN_ID && error.status && error.status >= 500) {
     return CommonErrorTypesConstants.APP_LEVEL_ERROR_APPOINTMENTS
+  } else if (!error.json?.errors[0].refreshable && error.status && error.status === 418) {
```
dumathane, line 47:
```suggestion
```

VAMobile/src/utils/errors.ts
```diff
@@ -44,6 +44,10 @@ export const getCommonErrorFromAPIError = (error: APIError, screenID?: ScreenIDT
...
+    return CommonErrorTypesConstants.CUSTOM_ERROR
```
dumathane, line 48:
```suggestion
```

VAMobile/src/utils/errors.ts
```diff
@@ -44,6 +44,10 @@ export const getCommonErrorFromAPIError = (error: APIError, screenID?: ScreenIDT
...
+  } else if (error.json?.errors[0].refreshable && error.status && error.status === 418) {
```
dumathane, line 49:
```suggestion
```

VAMobile/src/utils/errors.ts
```diff
@@ -44,6 +44,10 @@ export const getCommonErrorFromAPIError = (error: APIError, screenID?: ScreenIDT
...
+    return CommonErrorTypesConstants.CUSTOM_ERROR_WITH_REFRESH
```
dumathane, line 50:
```suggestion
```

dumathane: I may have mucked up the syntax playing in the web browser but something like this might be a little cleaner.

dumathane (changes requested):
This looks good overall but I have two comments, one needs a change, the other we could change:
1) I get why we are using 'CallHelpCenter' but given my 2nd issue I wonder if we should just make our own custom BE Error component that mirrors CallHelpCenter but is more specific to our wants/needs (as in doesn't default to anything that isn't provided for example). Like having legitimate a11y variables we could pass in and use as opposed to being forced to use them to make sure the text matches.

2) I think we need to code a way to enable a route that DOESNT have a phone number. Right now, we default to the main helpline but we've had AF situations crop up where a phone number doesn't make sense, so we want to be able to avoid having one if possible. This would mean a change to the CallHelpCenter component itself, or changing to a new component a la issue 1^.

VAMobile/src/utils/errors.ts
cadibemma: got it!

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { useTheme } from 'utils/hooks'
+import { fixedWhiteSpaceString } from 'utils/jsonFormatting'
+
+export type CustomErrorProps = {
+  /** optional function called when the Try again button is pressed */
+  onTryAgain?: () => void
+  /** optional text for the title */
+  titleText?: string
+  /** optional text for the error */
+  errorText?: string
```
dumathane, line 22: title and errortext shouldn't be optional, that seems like the bare minimum requirement.

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const scrollStyles: ViewStyle = {
+    justifyContent: 'center',
+  }
+
+  const containerStyles = {
+    mt: theme.dimensions.contentMarginTop,
+    mb: theme.dimensions.contentMarginBottom,
+  }
+  useEffect(() => {
+    logAnalyticsEvent(Events.vama_fail())
```
dumathane, line 43: We should make new analytics for this and not inherit the old one. Events.vama_be_af_shown ?

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, [])
+
+  const standardMarginBetween = theme.dimensions.standardMarginBetween
+
+  const tryAgain = () => {
+    logAnalyticsEvent(Events.vama_fail_refresh())
```
dumathane, line 49: Events.vama_be_af_refresh

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  return (
+    <VAScrollView contentContainerStyle={scrollStyles}>
+      <Box justifyContent="center" {...containerStyles}>
+        <AlertBox
+          title={fixedWhiteSpaceString(titleText)}
+          titleA11yLabel={titleA11y}
+          text={onTryAgain ? t('errors.callHelpCenter.sorryWithRefresh') : t('errors.callHelpCenter.sorry')}
```
dumathane, line 61: Drop this text in it's entirety I think.

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
```diff
@@ -0,0 +1,84 @@
...
+          border="error">
+          <Box>
+            <TextView
+              variant="MobileBody"
+              paragraphSpacing={true}
+              mt={theme.paragraphSpacing.spacing20FontSize}
```
dumathane, line 67: Drop this since we are killing the alertbox text.

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
```diff
@@ -0,0 +1,84 @@
...
+              accessibilityLabel={errorA11y}>
+              {fixedWhiteSpaceString(errorText)}
+            </TextView>
+            <ClickToCallPhoneNumber a11yLabel={a11yLabelID(callPhone)} phone={callPhone} />
```
dumathane, line 71: This needs to be optional on phone number existing I think.

dumathane:
A few more notes for Chika, so close!

VAMobile/src/components/CommonErrorComponents/CustomError.tsx
cadibemma: modified it to standard margin to maintain space text had from the title

dumathane (dismissed):
I think this look good. I think in retrospect we should have maybe opted for an informational alert type as a BE provided option as well but I don't think that's a show stopper. I'm going to ask Theo to review this code today too as well, make sure I'm not missing anything.

theodur (approved):

Sparowhawk (dismissed):

TKDickson:
Commented on ticket in more detail - this is exhibiting (in only Android that I've noticed) the same "get an error right away from the backend but then lag/freeze on front end" thing we've seen before with other errors.

TKDickson:
@theodur - @cadibemma just DM'd me that she has confirmed she can log in on both platforms. 

Approved for merge to develop; no QA done by QA.

TKDickson (approved):
approved to merge to develop

PR #8929, opened 2024-06-27
bug/8911-CountAfterCancellingAppointment

theodur:
## Description of Change

Cancelling an upcoming appointment within 7 days doesn't update the upcoming appointment count on the Home and Health screen. This is caused by the `cancelAppointment` mutation getting and setting data for the incorrect query key. After the changes from #8635, we're no longer fetching paginated pages from the backend, and therefore the query key for the `getAppointments` query was updated to **not** contain the page number. This change just needed to be made in the `cancelAppointment` mutation.

## Screenshots/Video

N/A because the appointments system won't let me cancel another appointment as of now.

## Testing

Tested by making an appointment for Judy and cancelling it.

- [ ] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

The upcoming appointment count is updated appropriately on the Home and Health screens after cancelling an appointment.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/api/appointments/cancelAppointment.tsx
+++ b/VAMobile/src/api/appointments/cancelAppointment.tsx
@@ -28,7 +28,7 @@ const cancelAppointment = (cancelID: string) => {
 /**
  * Returns a mutation for canceling an appointment
  */
-export const useCancelAppointment = (currentPage: number) => {
+export const useCancelAppointment = () => {
   const queryClient = useQueryClient()
 
   return useMutation({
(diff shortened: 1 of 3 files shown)


Sparowhawk (approved):

VAMobile/src/api/appointments/cancelAppointment.tsx
```diff
@@ -65,10 +64,7 @@ export const useCancelAppointment = (currentPage: number) => {
         },
       }
 
-      queryClient.setQueryData(
-        [appointmentsKeys.appointments, TimeFrameTypeConstants.UPCOMING, currentPage],
-        newAppointments,
-      )
+      queryClient.setQueryData([appointmentsKeys.appointments, TimeFrameTypeConstants.UPCOMING], newAppointments)
```
dumathane, line 67: This was the lone remaining api/ example that included 'currentPage' so I think we are safe saying this is the only lingering example.

TKDickson (approved):
approved to merge to feature branch

PR #9043, opened 2024-07-10
bug/8839-Chika-PaymentsQueryNotDependentOnAuthorizedServicesAPICall

cadibemma:
## Description of Change

- The `getPayments` query isn't blocked when the authorized services call returns the `paymentHistory` service as `false`. This should be accounted for in the query so the `/v0/payment-history` endpoint won't be called when the user isn't authorized for payment history. In this PR, the payments endpoint is blocked when user isn't authorized.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


Sparowhawk (approved):

theodur (changes requested):
The authorized checked should be applied to the query itself. You can use this pattern

https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/src/api/appointments/getAppointments.tsx#L44-L52

VAMobile/src/api/payments/getPayments.tsx
```diff
@@ -35,8 +37,11 @@ const getPayments = async (year: string | undefined, page: number): Promise<Paym
  * Returns a query for user payments history
  */
 export const usePayments = (year: string | undefined, page: number, options?: { enabled?: boolean }) => {
+  const { data: authorizedServices } = useAuthorizedServices()
+  const queryEnabled = options && has(options, 'enabled') ? options.enabled : true
   return useQuery({
     ...options,
+    enabled: !!(authorizedServices?.paymentHistory && queryEnabled),
```
theodur, line 44: This query should also be dependent on whether payments is in downtime

```tsx
const paymentsInDowntime = useDowntime(DowntimeFeatureTypeConstants.payments)
...
enabled: !!(authorizedServices?.paymentHistory && !paymentsInDowntime && queryEnabled),
```

theodur (changes requested):

VAMobile/src/api/payments/getPayments.tsx
cadibemma: Oh yeah i was going to ask whether downtime should be moved as well. Will make the change :+1:

TKDickson:
@cadibemma feel free to tell me no, but if you also happened to fix #9065 since you're already making changes to payment history calls, I would be totally OK with it :)

theodur:
@cadibemma I noticed you made the downtime additions to the payments query. Is this one ready for review again?

@TKDickson During my investigation of #9064, I noticed the bug you described in #9065. I think the two issues _may_ be related, because the reason the query, and ultimately the API call is being made twice is because the screen is being re-rendered unexpectedly, which is related to the error we're seeing in #9064. Chika's changes don't seem to fix this, so we'll probably want to investigate #9065 separately, probably alongside #9065

theodur (approved):

TKDickson (approved):
approved to merge to develop

Issue #8891, opened 2024-06-24
BUG - sev-3 - Fix areas with too much spacing in the app


brea11y:
## What happened?

QA noticed that there are areas in the app that are receiving too much spacing after paragraphs. Ticket #8799 was opened and some exploratory work was done. It was discovered that not only do we have too much spacing, but that our spacing is very inconsistent.

### Brea's findings:

> After doing some research, we are definitely inconsistent in our paragraph spacing. 
> 
> Initially, I took some time to review the WCAG guidance and additional guidance from other sources for accessibility best practices in native mobile apps (including MCAG). All sources to recommend line-height at 1.5x the font size and 2x the font size for paragraph spacing, which is what was recommended during our WCAG updates. However, this does not seem to be what we are doing throughout the app (likely due to new components being implemented) and we are seeing areas (like in the example above), with inconsistent spacing and too much spacing between elements.
> 
> I did some [exploration in Figma](https://www.figma.com/design/BVfWDc8vinwx8kRc1LoqGl/%F0%9F%8F%96-Brea's-Sandbox?node-id=1296-883&t=9HWuO4tZoiKpTuFQ-0) and discovered that many of the elements on the screens shown in the example have as much as 67px between elements, with the average amount of spacing being 64px. This should be 40px and should not increase even if links are present.
> 
> I [included a side-by-side comparison](https://www.figma.com/design/BVfWDc8vinwx8kRc1LoqGl/%F0%9F%8F%96-Brea's-Sandbox?node-id=1297-4748&t=9HWuO4tZoiKpTuFQ-4) in the Figma file which shows what is in the app today vs. what we should actually have in the app. I also took a look at some of the other areas of the app (mostly focusing in on the health area where we have longer blocks of copy) and included [some examples of what the spacing should be](https://www.figma.com/design/BVfWDc8vinwx8kRc1LoqGl/%F0%9F%8F%96-Brea's-Sandbox?node-id=1297-5334&t=9HWuO4tZoiKpTuFQ-4). 
> 
> I've talked to @wavelaurenrussell quite a bit about the spacing inconsistencies and I know that this is something that she was looking to resolve in her recent work on appointments, but this is really an app-wide issue that we need to resolve. 
> 
[...]

PR #9115, opened 2024-07-16
fixed

Sparowhawk:
## Description of Change
Was found that SM save message alert and appeal issues had incorrect applied paragraph spacing, fixed component props and then added paragraph spacing that needed to be re-added to all the other lists since they needed to maintain paragraph spacing

## Screenshots/Video
Screenshot 2024-07-16 at 2 47 16 PM
Screenshot 2024-07-16 at 2 47 31 PM

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
 think this was brought on by WCAG paragraph spacing but maybe not. There is a too much spacing after a bulleted list in some cases. This includes:

Messages > The we need more information bullet list when saving a messaging, saving a reply, saving a draft etc.
Claims > Claims History > Appeal Details > Issues Segment > The currently on appeal bullet list

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/auth/LoaGate/LoaGate.tsx
+++ b/VAMobile/src/screens/auth/LoaGate/LoaGate.tsx
@@ -66,7 +66,7 @@ function LoaGate({}: LoaGateProps) {
             <TextView {...bodyTextProps}>{t('loaGate.readMore.itemTwo.OfferProof')}</TextView>
           </Box>
           <Box mt={theme.dimensions.standardMarginBetween}>
-            <VABulletList listOfText={[bulletOne, { text: t('loaGate.readMore.bulletTwo') }]} />
+            <VABulletList listOfText={[bulletOne, { text: t('loaGate.readMore.bulletTwo') }]} paragraphSpacing={true} />
           </Box>
         </CollapsibleView>
         <Box mt={theme.dimensions.textAndButtonLargeMargin}>
--- a/VAMobile/src/components/VABulletList.tsx
+++ b/VAMobile/src/components/VABulletList.tsx
@@ -51,7 +51,7 @@ export type VABulletListProps = {
 /**
  * Displays the list of text as a bulleted list
  */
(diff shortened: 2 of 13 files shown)


alexandec (dismissed):

rbontrager:
I didn't realize the number of instances we had paragraph spacing in the app.  From Brea: 'I don’t believe we implemented the extra paragraph spacing that WCAG requires into the lists, so I think this might be a bug.' @Sparowhawk It sounds like we might just need the normal spacing before and after lists.

VAMobile/src/screens/auth/LoaGate/LoaGate.tsx
```diff
@@ -66,7 +66,7 @@ function LoaGate({}: LoaGateProps) {
             <TextView {...bodyTextProps}>{t('loaGate.readMore.itemTwo.OfferProof')}</TextView>
           </Box>
           <Box mt={theme.dimensions.standardMarginBetween}>
-            <VABulletList listOfText={[bulletOne, { text: t('loaGate.readMore.bulletTwo') }]} />
+            <VABulletList listOfText={[bulletOne, { text: t('loaGate.readMore.bulletTwo') }]} paragraphSpacing={true} />
```
theodur, line 69: I'm working on a similar [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/8891) for fixing the padding on Global related screens, and I had to make the same change you made [here](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/9115/files#diff-580f6e35c23c01211dbeffcd729de2a8cb34ec1eaec152aef1a6c67474b5d84dR54) on my branch. The difference is we don't need `paragraphSpacing` to be `true` though, because we want the padding to removed on the `LoaGate` screen, so this one can be removed

Sparowhawk: updated

alexandec (dismissed):

alexandec (approved):

rbontrager (approved):

PR #9141, opened 2024-07-19
Migrated play core

timwright12:
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/android/app/build.gradle
+++ b/VAMobile/android/app/build.gradle
@@ -236,8 +236,21 @@ dependencies {
     implementation("com.google.firebase:firebase-config")
 
     // playstore core for in-app reviews
-    implementation 'com.google.android.play:core:1.10.0'
-    implementation 'com.google.android.play:core-ktx:1.8.1'
+    // implementation 'com.google.android.play:core:1.10.0'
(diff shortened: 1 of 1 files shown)


alexandec:
The app built fine for me locally. I ran it in the Android simulator and played with it in demo mode. Didn't see any issues

alexandec (approved):

theodur (commented):
@timwright12 Just wanna make sure, did you also confirm that in-app reviews still works?

timwright12:
@theodur no, that's actually something I was curious about and not sure how to trigger it

theodur:
> @theodur no, that's actually something I was curious about and not sure how to trigger it

You have to be logged in to the Google Play Store on the device, and then you need to do a triggering event in the app at least 7 times to get past the review count threshold. The easiest way is probably to navigate to and back from the Personal information screen 7 times

timwright12:
@theodur thanks, I'll kick that over to QA

timwright12:
@theodur checked in app review and it still works. Thanks! Great catch there

PR #9191, opened 2024-07-25
bug/9187-SMEditDraftNullCrashesFix

Sparowhawk:
## Description of Change
Added a fix for recipientID being null as a result of 502 and 504 errors from upstream. Not sure why that caused the crash, but the only toString function that didn't have a ? after recipientID is in the useEffect and would cause it to possibly crash. Not sure how this can officially be tested though

## Screenshots/Video
N/A

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Shouldn't crash with 502s or 504s on the edit draft screen

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HealthScreen/SecureMessaging/EditDraft/EditDraft.tsx
+++ b/VAMobile/src/screens/HealthScreen/SecureMessaging/EditDraft/EditDraft.tsx
@@ -150,7 +150,7 @@ function EditDraft({ navigation, route }: EditDraftProps) {
   )
   const replyDisabled = isReplyDraft && !hasRecentMessages
 
-  const [to, setTo] = useState(message?.recipientId?.toString() || '')
+  const [to, setTo] = useState((message?.recipientId || '').toString())
   const [category, setCategory] = useState<CategoryTypes>(message?.category || '')
   const [subject, setSubject] = useState(message?.subject || '')
   const [attachmentsList, addAttachment, removeAttachment] = useAttachments()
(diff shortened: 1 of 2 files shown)


VAMobile/src/screens/HealthScreen/SecureMessaging/EditDraft/EditDraft.tsx
```diff
@@ -183,7 +183,7 @@ function EditDraft({ navigation, route }: EditDraftProps) {
       setBody(message?.body || '')
       setCategory(message?.category || '')
       setSubject(message?.subject || '')
-      setTo(message?.recipientId.toString() || '')
+      setTo(message?.recipientId?.toString() || '')
```
theodur, line 186: This would still cause an error if `recipientId` is `undefined` since `toString` expects an instance of some type. You'll probably need to do something like:
```tsx
setTo((message?.recipientId || '').toString())
```

The [type](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/534d62d4b7961123d7b32bec2ca3b9776252e6af/VAMobile/src/api/types/SecureMessagingData.ts#L23) for this attribute should also be updated to `recipientId?: number` so that TypeScript will throw an error if this is referenced anywhere else without caution

theodur (changes requested):

VAMobile/src/screens/HealthScreen/SecureMessaging/EditDraft/EditDraft.tsx
Sparowhawk: good catch and applied

theodur (approved):
Updates look good. I notice there's an SM E2E test that's failing, not sure whether it's due to your updates or not

rbontrager (approved):

PR #9222, opened 2024-07-30
bug/9119-Chika-NotLoggingVamaLoginTokenFetchErrors

cadibemma:
## Description of Change

- We have not logged a single `vama_login_token_fetch` error in production. It seems far more likely that our assumptions/logic behind how we could log that error are broken
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

- [ ] We should be logging those errors, with whatever related information would also be helpful, in production
## PR Checklist

## For QA


theodur (changes requested):
I noticed the logic for the `vama_login_token_fetch` event wasn't updated. That one will need the same changes you made to `vama_login_token_refresh`.

dumathane:
I think we can also just retire vama_exchange_failed as well.

VAMobile/src/constants/analytics.ts
```diff
@@ -761,11 +761,11 @@ export const Events = {
       },
     }
   },
-  vama_login_token_refresh: (status_code: number): Event => {
+  vama_login_token_refresh: (error: Error): Event => {
     return {
       name: 'vama_login_token_refresh',
       params: {
-        status_code,
+        error,
```
theodur, line 768: I believe the error also needs to be converted to a string

cadibemma: converted!

theodur (approved):

TKDickson:
@theodur @cadibemma have either/both of you built the app in both platforms with these changes and logged in? If not, I can add to my list (that's the basic stuff I want to do with all analytics changes, especially something that we'd be trying to fit into the last day of a sprint).

theodur:
@TKDickson I personally haven't. Not sure if Chika has, but I can do a quick test

theodur:
Confirmed login works on the iOS simulator

TKDickson:
@theodur - @cadibemma also just DM'd me that she has confirmed she can log in on both platforms. 

Approved for merge to develop; no QA done by QA.

(Also I originally put this comment on the wrong PR - the backend custom errors one - in case anyone got an email about that 🤦‍♀️ )

TKDickson (approved):

PR #9238, opened 2024-07-31
BUG - Version Mismatch Between ChromeDriver and Installed Chrome Browser

IsraelleHub:
## Description of Change
The error we are encountering is due to a version mismatch between the ChromeDriver and the installed Chrome browser. 
We need to install a ChromeDriver version that matches our Chrome browser version (126.0.6478.182) and specify its path when running the axe command.
[Error Link](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/10177268188)

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


.github/workflows/documentation_accessibility_checks.yml
```diff
@@ -66,19 +71,11 @@ jobs:
 
   start_slack_thread:
     name: Start Slack thread
-    runs-on: ubuntu-latest
     if: ${{ failure() }}
     needs: axe-scan
-    steps:
```
timwright12, line 72: Is this OK to remove?

timwright12 (changes requested):

.github/workflows/documentation_accessibility_checks.yml
```diff
@@ -7,6 +7,9 @@ on:
   # Run on Weekdays at 9:00 AM UTC, 4AM ET, 1:00 AM PT
   schedule:
     - cron: '0 9 * * 1,2,3,4,5'
+  push:
+    branches:
+      [bug-axe-check-chome-driver-version]
```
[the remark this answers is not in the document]
IsraelleHub: Yes, I am waiting for final review to remove it.

.github/workflows/documentation_accessibility_checks.yml
IsraelleHub: Yes. Now it is using start_slack_thread.yml instead. It is ok. Other Workflow are using the same format as well.

timwright12: Can you break this work off into another PR if it isn't related to the chromium issue?

.github/workflows/documentation_accessibility_checks.yml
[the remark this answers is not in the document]
timwright12: My mistake, I thought I read that this was ready for review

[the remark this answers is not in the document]
IsraelleHub: It is ready for review. it's ready for review

.github/workflows/documentation_accessibility_checks.yml
IsraelleHub: yes, I rolled it back

timwright12 (approved):

PR #9251, opened 2024-08-02
feature/9221-Binny-UpdateInAppReviewEventLocations

dumathane:
## Description of Change

Screen views and even detailed information views don't really work for these, or else we stumble into the 'Ameet airline effect' (being asked to leave app review in the middle or near the end of completing a task in the app). 

We discussed and we're going to move forward and circle back to the screen problem with [9253](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9253).

Technical investigation: I thought maybe we could useBeforeNavBackListener, similar to when we leave edited information on a screen, but those all inherently require to be on a screen that allows data entry which is setup differently on the nav stacks...
The problem is even if we setup useBeforeNavBackListeners (Health has some strange stack setup to allow for navigation to be passed), it still won't register moving to different tab nav stacks as an 'away'.

additional considerations:
downloading/viewing a file attachment in secure messaging?

There are essentially 4 types of reviewEvent register types:
1) On successful api request (request refill, edit email, update preferred name)
2) On dismissing a letter download. (One is on the other ticket)
3) useEffect on a screen load with details (to be fixed with above ticket).
4) one on screen dismiss as the only place we could do so due to it being a largePanel (Veteran Status)

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/api/contactInformation/deleteEmail.tsx
+++ b/VAMobile/src/api/contactInformation/deleteEmail.tsx
@@ -5,6 +5,7 @@ import { Events, UserAnalytics } from 'constants/analytics'
 import { Params as APIParams, EditResponseData, del } from 'store/api'
 import { logAnalyticsEvent, logNonFatalErrorToFirebase, setAnalyticsUserProperty } from 'utils/analytics'
 import { isErrorObject } from 'utils/common'
+import { registerReviewEvent } from 'utils/inAppReviews'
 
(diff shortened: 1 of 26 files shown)


cadibemma (approved):

TKDickson (approved):
approved to merge to develop

PR #9305, opened 2024-08-09
bug/9266-JSONParseError

theodur:
## Description of Change

There has been an increase in `vama_error` events, specifically containing the `JSON Parse error`. This is caused by the `/v0/user/logged-in` endpoint returning a `Content-Type` of `text/html`, which throws an error when `response.json` attempts to parse it as JSON.

To resolve this, I updated our fetch client function to return early if the response type is anything other than `application/json`, this way, non JSON responses won't be parsed. Ideally, the API shouldn't be returning a non JSON response, but this serves as a guard for whenever it does. 

## Screenshots/Video

N/A

## Testing

Tested by signing in to staging in the simulator. This probably can't be tested on a test build since the error wasn't crashing the app, and needs to be seen in the logs.

- [x] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

`vama_error` isn't logged for the `/v0/user/logged-in` endpoint when the response is successful.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/store/api/api.test.ts
+++ b/VAMobile/src/store/api/api.test.ts
@@ -82,13 +82,6 @@ context('api', () => {
     fetch.mockResolvedValue(response)
     const formData = new FormData()
 
-    it("returns undefined when the Content-Type of a successful response is not 'application/json'", async () => {
-      const responseBlob = new Blob(['Success'])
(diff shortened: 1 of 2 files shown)


dumathane (approved):

TKDickson (dismissed):
approved to merge to develop

TKDickson (approved):

PR #9381, opened 2024-08-15
bug/9366-rachael-detox-fix-nightly-build

rbontrager:
## Description of Change

Claims and decision letters were failing in iOS and navigation in Android.  Passing iOS test case (minus those cases affected by in app review that we are removing): [iOS](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/10414462651).  Passing android test cases [here](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/10421135675)

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/.detoxrc.json
```diff
@@ -28,7 +28,7 @@
       "type": "ios.simulator",
       "device": {
         "type": "iPhone 15 Pro Max",
-        "os": "iOS 17.4"
+        "os": "iOS 17.2"
```
TKDickson, line 31: Did you intend to include this change?

TKDickson (approved):

VAMobile/.detoxrc.json
rbontrager: Just so there is a record.  Yes I did because the detox runs are still usually getting 1 or 2 timeouts for iOS.  I changed it to see if this works better.  I didn't have any timeouts on the detox run I did so I'm hopeful this will remove them all together.

PR #9412, opened 2024-08-20
feature/9286-SingleSignOn

theodur:
## Description of Change

Integrates SSO into the app and adds functionality to authenticate WebViews. This is currently placed behind the `sso` feature toggle. In the PR, the only feature that takes advantage of SSO is the scheduling appointments link. I left this in there so that SSO functionality could be tested by QA, but depending on which feature we decide to use SSO on first, that part of the PR may change.

## Screenshots/Video

https://github.com/user-attachments/assets/80f8bf14-502d-4a0e-b2c2-f0fa4185826f

## Testing

Tested on staging with Judy and manually forcing the `NoAppointments` component to show. This will either need to be tested with a user that doesn't have any appointments or the `/v0/appointments` endpoint will need to be mocked with Charles Proxy to return an empty appointments array.

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/env/env.sh
```diff
@@ -50,6 +50,10 @@ else
 fi
 # set demo mode password
 echo "DEMO_PASSWORD=${DEMO_PASSWORD}" >> .env
+
+# set website URLs
+echo "LINK_URL_SCHEDULE_APPOINTMENTS=https://${WEBSITE_PREFIX}va.gov/health-care/schedule-view-va-appointments" >> .env
+
```
dumathane, line 56: What is the functional purpose of this that i'm missing?

dumathane (approved):
Approved, but answer the .env question for me if you could.

VAMobile/env/env.sh
theodur: `LINK_URL_SCHEDULE_APPOINTMENTS` was hard-coded as https://www.va.gov/health-care/schedule-view-va-appointments/ in `constants.env`, so even on a staging build, it would take you to the prod site because it was static. This change makes the link dynamic, so it'll adjust the URL based on the environment, which allows us to test in staging. Future links that we want to use SSO with will need to be added to the `env.sh` script similarly

dumathane (dismissed):

VAMobile/src/screens/WebviewScreen/WebviewScreen.tsx
```diff
@@ -155,6 +157,14 @@ function WebviewScreen({ navigation, route }: WebviewScreenProps) {
...
         const cookieHeaders = response.headers.get('set-cookie')
         cookieHeaders && (await CookieManager.setFromResponse(AUTH_SIS_TOKEN_EXCHANGE_URL, cookieHeaders))
+
+        const cookies = await CookieManager.get(AUTH_SIS_TOKEN_EXCHANGE_URL)
+        const cookiesArray = Object.values(cookies)
+        const hasSSOCookies = SSO_COOKIE_NAMES.every((cookieName) =>
+          cookiesArray.some((cookie) => cookie.name === cookieName),
+        )
+
+        logAnalyticsEvent(Events.vama_sso_cookie_received(hasSSOCookies))
```
dumathane, line 192: Maybe we should log the screen we're going to as well?

dumathane (approved):

TKDickson:
@theodur this appears to be more broken than the original problem we were trying to solve, unfortunately - sending back for changes with comment in ticket.

VAMobile/src/screens/WebviewScreen/WebviewScreen.tsx
theodur: SSO cookie retrieval failing won't be dependent on the URL they're visiting, but it can't hurt to have that extra data. Just added that in

TKDickson (approved):

cadibemma (approved):
Looks good!

PR #9429, opened 2024-08-21
bug/9325-ClaimsLoadingOnReviewFileRequests

Sparowhawk:
## Description of Change
passed the already loaded claim to the file request screen, no need to reload due to the refresh timer right after a user is already on the details screen. Does fall back on api if we need to load for whatever reason of the claim data being undefined.

## Screenshots/Video

https://github.com/user-attachments/assets/46b280e7-5d82-4a9d-b4a4-604d4dc053a5

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
no longer displays unnecessary loading screens

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/FileRequest.tsx
+++ b/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/FileRequest.tsx
@@ -33,12 +33,21 @@ function FileRequest({ navigation, route }: FileRequestProps) {
   const theme = useTheme()
   const { t } = useTranslation(NAMESPACE.COMMON)
   const navigateTo = useRouteNavigation()
-  const { claimID } = route.params
-  const { data: claim, error: claimError, refetch: refetchClaim, isFetching: loadingClaim } = useClaim(claimID)
-  const requests = currentRequestsForVet(claim?.attributes.eventsTimeline || [])
+  const { claimID, claim } = route.params
+  const {
+    data: claimFallBack,
+    error: claimError,
+    refetch: refetchClaim,
+    isFetching: loadingClaim,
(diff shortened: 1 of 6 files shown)


VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/FileRequest.tsx
```diff
@@ -33,12 +33,21 @@ function FileRequest({ navigation, route }: FileRequestProps) {
   const theme = useTheme()
   const { t } = useTranslation(NAMESPACE.COMMON)
   const navigateTo = useRouteNavigation()
-  const { claimID } = route.params
-  const { data: claim, error: claimError, refetch: refetchClaim, isFetching: loadingClaim } = useClaim(claimID)
-  const requests = currentRequestsForVet(claim?.attributes.eventsTimeline || [])
+  const { claimID, claim } = route.params
```
alexandec, line 36: Seems like passing the claim as a prop is a reasonable solution. But shouldn't React Query be caching the claim to prevent this issue in the first place? If the user just visited the parent screen which fetched and cached the claim, why does fetching the same claim in the child component cause the loading indicator to appear? Caching should allow fetching the claim in the child component without having to pass it as a prop. Just trying to understand the root cause better

alexandec (approved):
Approving but also added a question

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/FileRequest.tsx
Sparowhawk: it's because when this occurs is when the stale time is up, it isn't always here, but it does occur regularly enough that we should just pass the claim further down anyways

rbontrager (approved):

PR #9447, opened 2024-08-22
Automation/9431-chanel-automation-make-release-pr-branch-unique

IsraelleHub:
## Description of Change
Our [release_pull_requests workflow](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/.github/workflows/release_pull_request.yml) creates PRs to merge any changes that were made on the release/vX.X.X branch during the RC phase back into develop, and then merges release/vX.X.X to the main branch.

If the PR that merges changes back to develop is not merged and closed before the next release, the dev-to-release branch remains open and the next release build will fail since the old dev-to-release branch already exists. We should make these merge branch names unique to prevent this in the future.

[Example failure](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/10480125161/attempts/1)

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Acceptance Criteria
- [ ] Replace the creation of new branch dev-to-release in [release_pull_requests](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/a471fce31ea9098dad888e7ee4d4a7509746f727/.github/workflows/release_pull_request.yml#L47) to be a unique and more accurate name such as release-vX.X.X-to-develop
 - [ ] Add the slack failure notification to approve_command.yml

## PR Checklist

## For QA

### DIFF
--- a/.github/workflows/release_pull_request.yml
+++ b/.github/workflows/release_pull_request.yml
@@ -44,8 +44,8 @@ jobs:
           git config --global user.name 'VA Automation Bot'
           git config --global user.email 'va-mobileapp@adhocteam.us'
           git checkout release/${{inputs.version}}
-          git checkout -b release- ${{inputs.version}}-to-develop
-          git push -u origin release- ${{inputs.version}}-to-develop
+          git checkout -b release-${{inputs.version}}-to-develop
+          git push -u origin release-${{inputs.version}}-to-develop
           echo "DEV_PR=$(gh pr create -B develop -t "Merge ${{inputs.version}} to develop" -b "PR to merge ${{inputs.version}} release branch changes into develop")" >> $GITHUB_ENV
       - name: 'Merge changes to main and tag for release build'
         run: |
@@ -53,7 +53,7 @@ jobs:
           git config --global user.email 'va-mobileapp@adhocteam.us'
           git checkout main
           git checkout release/${{inputs.version}}
-          git checkout -b merge-to-main
+          git checkout -b merge-${{inputs.version}}-to-main
           git merge -s ours main -m "merge main changes"
           git checkout main
           git merge merge-to-main --squash
--- a/.github/workflows/release_pull_request.yml
+++ b/.github/workflows/release_pull_request.yml
@@ -30,10 +30,6 @@ jobs:
       devPrUrl: ${{steps.urls.outputs.dev}}
       releaseHash: ${{steps.urls.outputs.release}}
     steps:
-      - name: 'Define branch variables'
-        run: |
-          echo "MERGE_TO_DEV_BRANCH=release-${{inputs.version}}-to-develop" >> $GITHUB_ENV
-          echo "MERGE_TO_MAIN_BRANCH=merge-${{inputs.version}}-to-main" >> $GITHUB_ENV  
       - name: 'Checkout repo'
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release_pull_request.yml
```diff
@@ -44,16 +44,16 @@ jobs:
...
         run: |
           git config --global user.name 'VA Automation Bot'
           git config --global user.email 'va-mobileapp@adhocteam.us'
           git checkout main
           git checkout release/${{inputs.version}}
-          git checkout -b merge-to-main
+          git checkout -b merge-${{inputs.version}}-to-main
           git merge -s ours main -m "merge main changes"
           git checkout main
           git merge merge-to-main --squash
```
narin, line 59: `merge-to-main` needs to be replaced on this line as well.

.github/workflows/release_pull_request.yml
```diff
@@ -44,16 +44,16 @@ jobs:
           git config --global user.name 'VA Automation Bot'
           git config --global user.email 'va-mobileapp@adhocteam.us'
           git checkout release/${{inputs.version}}
-          git checkout -b dev-to-release
-          git push -u origin dev-to-release
+          git checkout -b release-${{inputs.version}}-to-develop
```
narin, line 47: Since this branch name is used more than once, consider defining a variable above, i.e.

`MERGE_TO_DEV_BRANCH="release-${{inputs.version}}-to-develop"`

Then use it on lines 47 and 48 in the `git` commands. This will allow us to only change it in one spot if we need to in the future, and avoid any potential typos from repeating the expression.

The same goes for `merge-${{inputs.version}}-to-main`.

narin (changes requested):
Looking good but missed one branch name. Also suggesting you assign the branch names to variables since they are referenced more than once.

.github/workflows/release_pull_request.yml
```diff
@@ -39,24 +43,25 @@ jobs:
         run: echo "${{secrets.GH_ACTIONS_PAT}}" >> token.txt
       - name: 'Log into GH CLI'
         run: gh auth login --with-token < token.txt
+
```
narin, line 46: Unneeded whitespace

.github/workflows/release_pull_request.yml
```diff
@@ -30,6 +30,10 @@ jobs:
       devPrUrl: ${{steps.urls.outputs.dev}}
       releaseHash: ${{steps.urls.outputs.release}}
     steps:
+      - name: 'Define branch variables'
+        run: |
+          echo "MERGE_TO_DEV_BRANCH=release-${{inputs.version}}-to-develop" >> $GITHUB_ENV
+          echo "MERGE_TO_MAIN_BRANCH=merge-${{inputs.version}}-to-main" >> $GITHUB_ENV
```
narin, line 36: Is there a reason why this is its own step and they are set as workflow env vars instead of just normal bash variables within each of the steps? 

Workflow vars are typically used when they need to be used across multiple steps. These branch variables are only used within their own step so you probably don't need to define them at the workflow env level. See the following example in [release_branch_issue.yml](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/4f29195968a96a371a5726b7b8402638b3fd1913/.github/workflows/release_branch_issue.yml#L70)

narin (changes requested):
Suggestion on defining the variables but this is very close.

narin (approved):

PR #9491, opened 2024-08-28
bug/9307-EventTimelineNullSafety

Sparowhawk:
## Description of Change
Added null checking to the instances of eventTimeline that wasn't there previously to safeguard against it being null.

## Screenshots/Video
N/A

## Testing
yarn test, tested locally with console commands to make sure this would work if event timeline was null

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
it no longer crashes on null event timeline

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimTimeline/DEPRECATED_ClaimPhase.tsx
+++ b/VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimTimeline/DEPRECATED_ClaimPhase.tsx
@@ -105,7 +105,7 @@ function DEPRECATED_ClaimPhase({ phase, current, attributes, claimID }: ClaimPha
     testID = `${testID} ${updatedLastDate}`
   }
 
-  const count = numberOfItemsNeedingAttentionFromVet(eventsTimeline)
+  const count = numberOfItemsNeedingAttentionFromVet(eventsTimeline || [])
 
   const detailsText = t(`claimPhase.5step.details.phase${phase}`)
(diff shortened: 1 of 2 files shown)


VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimTimeline/DEPRECATED_ClaimPhase.tsx
```diff
@@ -105,7 +105,7 @@ function DEPRECATED_ClaimPhase({ phase, current, attributes, claimID }: ClaimPha
     testID = `${testID} ${updatedLastDate}`
   }
 
-  const count = numberOfItemsNeedingAttentionFromVet(eventsTimeline)
+  const count = numberOfItemsNeedingAttentionFromVet(eventsTimeline || [])
```
alexandec, line 108: Consider updating the function itself to accept an undefined or null value. That way the callers don't have to worry about passing particular arguments

alexandec (approved):
Approving with an optional suggestion

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimTimeline/DEPRECATED_ClaimPhase.tsx
Sparowhawk: noted, I believe it should remain at the screen/component level that follows our other patterns so far where we null check at the screen level and fall back on a default if we have it.

rbontrager (approved):

PR #9490, opened 2024-08-28
[Docs] Move Design System Changelog

narin:
## Description of Change

Docusaurus is set up to copy certain markdown files from the va-mobile-library repo. We are changing the location of the changelog so the file copy config needs to be updated.

Corresponding `va-mobile-library` PR: https://github.com/department-of-veterans-affairs/va-mobile-library/pull/457

## Screenshots/Video

N/A

## Testing

Tested changelog script locally

## PR Checklist

  **Reviewer:** Confirm the items below as you review

- [ ] PR is connected to issue(s)
- [ ] Tests are included to cover this change (when possible)
- [ ] No magic strings (All string unions follow the [Union -> Constant](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/src/constants/common.ts) type pattern)
- [ ] No secrets or API keys are checked in
- [ ] All imports are absolute (no relative imports)
- [ ] New functions and Redux work have proper TSDoc annotations

## For QA

[Run a build for this branch](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/workflows/on_demand_build.yml)

PR #9500, opened 2024-08-28
[Docs] Fix sidebar title for DS changelog

narin:
## Description of Change

The [last PR](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9490) updated the changelog location but had the incorrect link title in the sidebar. This PR fixes the link title. 

### Before/After
&nbsp;&nbsp;

## Testing

Local testing

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/documentation/docusaurus.config.js
+++ b/VAMobile/documentation/docusaurus.config.js
@@ -27,6 +27,8 @@ const engineeringDocForm = (name, URLextension, documentsList) => {
             header = `---\nsidebar_position: 1\n---\n\n`
             break
           case 'CHANGELOG.md':
+            filename = 'changelog.md'
+            break
           case 'README.md':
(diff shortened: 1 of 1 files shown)


TimRoe (dismissed):
Approving, but in trying to recall why I did things the way I did I think there's a "better" solution of leaving it in the 'documentation' section and updating the switch-case for `CHANGELOG.md` to:
```
filename = 'changelog.md'
break
```
if I am recalling correctly there was two issues with Changelog:
1. The all caps name screwed up it default sorting by file name so the switch-case change above would fix that
2. Way back it didn't have the "Changelog" as a header in the markdown itself so it needed a title added manually like the README's, but it seems like it does now so I think it should correctly set the sidebar name based on that without needing an override

TimRoe (approved):

Issue #9190, opened 2024-07-25
Add Page to Docusaurus - Updating Existing Get Started Content & Add Using Codespace 


bischoffa:
**Description:** 
As VAHB Mobile, I want to add documentation to the Docusaurus Site that shows/tells people how to use Codespaces to modify the site and add content. 

**Notes:** 

- [Slack Thread 7/25](https://adhoc.slack.com/archives/C02N65QRRB7/p1721854898750539) 
- Want to share this content / material so we do not use the old way of making changes or pull FE from projects to make these changes 
- Tim created  [imports pages of documentation from our separate repo](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/documentation/docusaurus.config.js#L7-L69) which is used for Design System 

**ACs:** 

- Update the dedicated page on how to use Docusaurus to include Codespaces 
- Remove content that is no longer relevant or needed
- If available / possible add tutorial videos

PR #9511, opened 2024-08-29
Doc-Removing misleading instructions while awaiting later ticket

TKDickson:
## Description of Change

Non-engineers who need to update the doc site had very specific instructions on the doc site about how to do so (that had them set up a bunch of stuff, make changes locally, then push them, etc).

For their ease of working, they should now be doing that via Codespaces. We don't currently have documentation for that (ticket #9190), but while we're awaiting that being prioritized we can at least _not_ tell people to do the old, slow, broken way.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


timwright12 (changes requested):
I'm good with added the new part, but that doesn't invalidate the old part. We'll still need it in there.

timwright12:
@TKDickson yeah it's still a valid way to update the docs site

timwright12 (approved):

Issue #9541, opened 2024-09-05
Create engineering documentation around demo mode


timwright12:
## Description

We need to support experience teams working in the app; part of that includes having them manage their own features in demo mode. We don't currently have sufficient documentation around adding and managing features in demo mode. This ticket supports doing that.

## Acceptance Criteria

- [ ] Add documentation to the main docs site for adding a feature into demo mode
- [ ] Add documentation to the main docs site managing demo mode data
- [x] Add general documentation about accessing demo mode

RamonAlvesPS:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

theodur:
The infrastructure around demo mode could likely change based on the discovery from this [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/6798). It doesn't look like any follow up tickets were made from that discovery, but I remember in FE sync, we (FE engineers) all agreed that integrating [msw](https://mswjs.io/) into the repo for demo mode seemed like a good idea. hat could potentially impact how soon we work on this

timwright12:
Demo mode documentation to migrate: https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/products/va-mobile-app/Mobile%20Demo%20Mode

theodur:
@timwright12 Is this one good to close now that the PR has been merged?

Issue #6798, opened 2023-09-21
Investigate React Query Demo Mode Options


theodur:
## Description

Currently demo mode is centered around the Redux store's `demoSlice`. This approach still works with React Query, since queries and mutations are ultimately making API calls that are being mocked and updated in the store, but it would be good to look at alternative solutions to see if we can remove the need for using the Redux store for async calls in demo mode. It could be that using the Redux store is still the best aproach, since the infrastructre is in place and already works well.

One alternative could be to use an API mocking library like [msw](https://github.com/mswjs/msw) to mock API requests.

## Acceptance criteria

- [ ] Investigate alternative ways of handling demo data with React Query
- [ ] Create appropriate implementation tickets if a better approach is discovered

Sparowhawk:
upon review of the library it is my recommendation that we migrate the json mocks over to using msw and not the redux store. It would help in the epic #8080 that is updating our demo mode data to break the mocks out into separate files for each api, it would also benefit us by updating our data. The added benefit of using msw is that the unit test json mocks will no longer be needed and detox and unit tests will be more aligned. 

I have included setting up msw and migrating over as a requirement in the 8080 epic which will then be broken down further to other tickets once we get to them.

Sparowhawk:
leaving this in code review until FE Sync to discuss with everyone else

PR #9544, opened 2024-09-05
feature/9348-DemoModeAPIOverrides

Sparowhawk:
## Description of Change
Added override api screen and setup each get/fetch api to check for override api and to throw the corresponding error supplied

## Screenshots/Video

https://github.com/user-attachments/assets/9f87db91-c502-43d0-8b9f-3ac5d5a30e78

## Testing
yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Currently there isn't an easily written detox solution for implementing errors from a backend. With this change, it will open us up to the availability of error testing in detox and make QA/VQA easier to manage if it's accessible through the developer menu.

## PR Checklist

## For QA


theodur (commented):
It's cool to see this in action! After digging into the PR, I noticed this approach requires all existing queries, and future queries to include the code snippet that iterates through the `overrideErrors` cache and throws an error if the query key exists there. Although this works, I'm not sure that this is the best pattern for mocking errors. 

Instead of having a new query for storing mocked errors and requiring all query functions to check for errors in the new error query cache, I think integrating MSW for demo mode would be a better route, and would avoid the need to change query functionality to accommodate for demo mode. It would also avoid using React Query in an antipattern way as a global synchronous storage for errors. Integrating MSW would require us to change our demo mode infrastructure, but we were already planning on doing that.

We had a [discovery ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/6798) to investigate using MSW, but it doesn't seem like any follow-up tickets were made for integrating it into the repo. Ironically, I made [one](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9580) earlier today based off of a [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9541) to add documentation for demo mode. Even though it's a much bigger lift that needs to be planned for, I feel like MSW should be integrated into the repo, and the functionality for mocking API calls to return errors would stem from that integration

Sparowhawk:
> It's cool to see this in action! After digging into the PR, I noticed this approach requires all existing queries, and future queries to include the code snippet that iterates through the `overrideErrors` cache and throws an error if the query key exists there. Although this works, I'm not sure that this is the best pattern for mocking errors.
> 
> Instead of having a new query for storing mocked errors and requiring all query functions to check for errors in the new error query cache, I think integrating MSW for demo mode would be a better route, and would avoid the need to change query functionality to accommodate for demo mode. It would also avoid using React Query in an antipattern way as a global synchronous storage for errors. Integrating MSW would require us to change our demo mode infrastructure, but we were already planning on doing that.
> 
> We had a [discovery ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/6798) to investigate using MSW, but it doesn't seem like any follow-up tickets were made for integrating it into the repo. Ironically, I made [one](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9580) earlier today based off of a [ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9541) to add documentation for demo mode. Even though it's a much bigger lift that needs to be planned for, I feel like MSW should be integrated into the repo, and the functionality for mocking API calls to return errors would stem from that integration

I agree. Ideally we would migrate over to MSW and I think we need to have that happen at some point. I do think this solution has some merit as a temporary work around until we migrate over to MSW.

theodur:
@Sparowhawk If we really want this in before the MSW migration, have you looked into using the demo store for this? Since we already use the demo store for mocking API calls, I'd imagine we could have an error state for each `get` endpoint that's used to determine whether it should return the mock data or an error. That would keep parity with the current demo mode infrastructure, instead of intertwining 2 different libraries to handle mocking demo mode responses. It would also be safer to use the demo store because it avoids the need to update query logic that's used in production for the sake of handling demo mode errors

Sparowhawk:
> @Sparowhawk If we really want this in before the MSW migration, have you looked into using the demo store for this? Since we already use the demo store for mocking API calls, I'd imagine we could have an error state for each `get` endpoint that's used to determine whether it should return the mock data or an error. That would keep parity with the current demo mode infrastructure, instead of intertwining 2 different libraries to handle mocking demo mode responses. It would also be safer to use the demo store because it avoids the need to update query logic that's used in production for the sake of handling demo mode errors

I did and this is what I found trying to do it that way.

1. needlessly difficult and not customizable(relying on json files of preset errors that wouldn't really be editable how we would like)
2. We would need to specify exactly which api's as opposed to query keys initially(which for claims/appeals/prescription tracking/messages would be fairly difficult for us to navigate when using the feature)
3. speaking more of the customizable issue it would limit us to basically known errors and responses that we have seen previously, where this solution is a little more dynamic and allows us to input more or less whatever we would like and would be fairly expandable in the future.
4. The other concern was time to implement here for immediate benefit as opposed to having to migrate to msw first.
5. Lastly all the demo mode store mocks have slightly different ways in how they are determined to be run so each demo mode store would need a slightly different solution as opposed to a universal solution which msw and this method provide.

dumathane:
I talked to Dylan and he is going to suggest a simpler route forward with this one.

Sparowhawk:
> I talked to Dylan and he is going to suggest a simpler route forward with this one.

I updated it to how I think it should best be handled at least at this juncture. @dumathane

Sparowhawk:
The updated way broke the unit tests badly that could not be fixed in the unit tests themselves from what I can tell and have tried. So reverting this back in a different branch

theodur (changes requested):
This is a big improvement from the last iteration. In regard to storing the error overrides, can we store those in the Redux store instead of using the Query Cache? We have the demo slice, which might be a good place to store this data. Maybe in an `errorOverides` state variable within that slice.

As mentioned in a previous comment, React Query isn't intended for storing synchronous data. I'm going to be working on documentation for React Query usage/best practices in the app soon, and this is one of the points that will be emphasized.

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
```diff
@@ -0,0 +1,551 @@
...
+          </Box>
+        </TextArea>
+        <TextArea>
+          <TextView accessibilityRole="header" variant="MobileBodyBold">
+            Authorized Services
+          </TextView>
+          <Box my={theme.dimensions.standardMarginBetween}>
+            {IndividualQueryDisplay('/v0/user/authorized-services', temporaryErrors, setErrors, clearData, t, theme)}
+          </Box>
+        </TextArea>
```
theodur, line 383: Instead of repeating the content for each endpoint, couldn't you make an array for the endpoints, and loop over that? e.g.

```tsx
const endpoints = [
  {
    name: 'Appointments',
    endpoint: '/v0/appointments',
  },
  {
    name: 'Authorized Services',
    endpoint: '/v0/user/authorized-services'
  },
  ...
]
```

This would result in less lines of code, and would also make it easier to add endpoints to this list in the future.

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
```diff
@@ -0,0 +1,551 @@
...
+import { useAppDispatch, useTheme } from 'utils/hooks'
+
+type OverrideAPIScreenProps = StackScreenProps<HomeStackParamList, 'OverrideAPI'>
+
+const IndividualQueryDisplay = (
+  endpoint: string,
+  overrideErrors: errors[],
+  setErrors: React.Dispatch<React.SetStateAction<errors[]>>,
+  clearErrors: boolean,
+  t: TFunction,
```
theodur, line 27: We don't need to pass the translation function since this is a component. It can be accessed directly with the `useTranslation` hook, which would make the usage of this component cleaner

theodur (changes requested):

VAMobile/src/api/types/ErrorData.ts
```diff
@@ -0,0 +1,30 @@
...
+      source: string
+      body?: string
+      telephone?: string
+      refreshable?: boolean
+      meta?: {
+        messages?: Array<ErrorMessage>
+      }
+    }>
+  }
+}
```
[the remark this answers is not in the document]
Sparowhawk: You are right. I can just use APIError. I removed this file and the index file for these types and rebased everything off the APIError Type

VAMobile/src/api/types/ErrorData.ts
```diff
@@ -0,0 +1,30 @@
+import { ErrorMessage } from 'store/api'
+
+export type errors = {
```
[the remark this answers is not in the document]
Sparowhawk: I removed this in favor of using APIError in the pre existing types

VAMobile/src/api/types/ErrorData.ts
```diff
@@ -0,0 +1,30 @@
...
+  endpoint: string
+  error: networkError | errorOverride
+}
```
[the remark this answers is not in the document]
Sparowhawk: I am now using the APIError type and rebased everything on that

VAMobile/src/store/api/api.ts
```diff
@@ -210,6 +210,24 @@ const call = async function <T>(
...
   } else {
+    const overrideErrors = _store?.getState().demo.overrideErrors
+    if (overrideErrors) {
+      _.forEach(overrideErrors, (error) => {
+        if (error.endpoint === endpoint) {
+          throw error.error
+        }
+        if (endpoint.includes(error.endpoint)) {
+          throw error.error
+        }
```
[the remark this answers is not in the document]
Sparowhawk: When implementing this I only took into account get/fetch calls and not any other types of calls.

So those 2 use cases wouldn't have been caught for the includes.

I agree the 2nd if statement is redundant and would catch the first so I removed the first if statement.

The 2nd statement with the includes is needed though because of fetching secure messages, claims, appeals, etc... that have a specific ID. Instead of it having to match exactly for each individual one I opted to match the service itself. So all messages will fail or all claims will fail and the same for everything else that has an ID or path parameter. This is also why the last check is needed is because the ${folderID} conflicts

VAMobile/src/store/api/api.ts
```diff
@@ -210,6 +210,24 @@ const call = async function <T>(
...
+        if (
+          error.endpoint === '/v0/messaging/health/folders/${folderID}/messages' &&
```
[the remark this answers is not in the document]
Sparowhawk: No this would not. The first part of this if is checking if the override error endpoint is that value. The other 2 parts are checking if the api called by our usage is /v0/messaging/health/folders/1/messages or a different folder. Both parts of this one is needed to not unintentionally throw another error for a different api.

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
Sparowhawk: I removed translation and theme being passed in and I am calling it in the component.

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
Sparowhawk: updated to an array of api groupings

VAMobile/src/store/slices/demoSlice.ts
```diff
@@ -26,6 +29,18 @@ export const updateDemoMode =
...
+    }
+  }
+
+/**
+ * sets the error overrides for demo mode
+ */
+export const updateErrorOverrides =
+  (errorOverrides: Array<errors>): AppThunk =>
+  async (dispatch) => {
+    dispatch(dispatchUpdateErrors(errorOverrides))
```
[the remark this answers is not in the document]
theodur: I think what you may have been missing is updating the demo slice to export its [actions](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/16292b34e5a1a78671b0601c5d2862aab31c8cc2/VAMobile/src/store/slices/demoSlice.ts#L61), e.g.

```tsx
export const { dispatchUpdateDemoMode, dispatchUpdateErrors } = demoSlice.actions
```

Then you should be able to call `dispatchUpdateErrors` like how I did in my earlier comment

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
```diff
@@ -0,0 +1,542 @@
...
+      {
+        endpoint: '/v0/appeal/',
+      },
+      {
+        endpoint: '/v0/claim/',
+      },
+      {
+        endpoint: '/v0/claims-and-appeals-overview',
+      },
+    ],
```
theodur, line 49: Since`endpoints` is essentially just an array of endpoints, could this be formatted to have the values stored directly in the array? e.g.

``` tsx
endpoints: [
  '/v0/appeal/',
  '/v0/claim/',
  '/v0/claims-and-appeals-overview',
]
```

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
```diff
@@ -0,0 +1,542 @@
...
+import { NAMESPACE } from 'constants/namespaces'
+import { HomeStackParamList } from 'screens/HomeScreen/HomeStackScreens'
+import { RootState } from 'store'
+import { APIError } from 'store/api'
+import { DemoState, updateErrorOverrides } from 'store/slices/demoSlice'
+import { useAppDispatch, useTheme } from 'utils/hooks'
+
+type OverrideAPIScreenProps = StackScreenProps<HomeStackParamList, 'OverrideAPI'>
+
+const APIGroupings = [
```
theodur, line 20: It would be helpful to create a type for this, so that when others developers add endpoints in the future, TypeScript would catch improper formatting

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
```diff
@@ -0,0 +1,542 @@
...
+      <TextArea>
+        <TextView accessibilityRole="header" variant="MobileBodyBold">
+          {group.name}
+        </TextView>
+        {individualQueries}
+      </TextArea>
+    )
+  })
+  return <Box>{groupings}</Box>
+}
```
theodur, line 224: Since this is only used in one place, and it's not much code, can this be moved directly into the screen component? A lot of the props for this component are being passed from the screen component anyway

VAMobile/src/store/slices/demoSlice.ts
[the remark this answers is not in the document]
Sparowhawk: That is what I missed. Thanks. This has been updated

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
Sparowhawk: updated, good call out

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
Sparowhawk: good call out

VAMobile/src/screens/HomeScreen/ProfileScreen/SettingsScreen/DeveloperScreen/OverrideApiScreen.tsx
Sparowhawk: I added a type to the APIGroupings to have TypeScript catch formatting

theodur (approved):

theodur (approved):

theodur (approved):

theodur (dismissed):

theodur (approved):

rbontrager (approved):

PR #9633, opened 2024-09-20
feature/8784-UpdateLogos

theodur:
## Description of Change

Replaces the current service emblems icons with the images provided by the component library's `mobile-asset` package. 

The remaining `Logo` instances are also replaced with the `VALogo` component. A new prop was added to `VALogo` for manually setting the color variant of the logo because certain screens like Veteran Status and the Onboarding screen need to it always be the same color.

This PR also adds functionality to dynamically use images from any directory without needing  to manually copy them to `android/app/src/main/res` or `ios/VAMobile/Images.xcassets`.

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


dumathane (dismissed):

narin:
Just chiming in to say you're awesome Theo. How'd you figure out this solution? Is it documented somewhere?

theodur:
> Just chiming in to say you're awesome Theo. How'd you figure out this solution? Is it documented somewhere?

@narin Thanks! 🙏 Through a lot of research, I found out that you're able to set a location for assets referenced in the build with the `--assets-dest` CLI option. Since assets need to be in the `android/app/src/main/res` directory for Android to be pull into the build, I set the assets-dest to that and it worked. 

iOS was ironically a little more tricky, I initially tried setting the assets-dest to `VAMobile/ios/VAMobile/Images.xcassets` since that's where we store our images for iOS, but that didn't seem to work. Then I discovered you can add a new location to the `Copy Bundle Resources` section in the project settings from the Xcode UI, which allows resources to be referenced when bundling the iOS build. I changed the assets-dest to be just the `ios` directory, which generates an `aseets` directory there with all the assets referenced in the code, and then I added that `assets` directory to the `Copy Bundle Resources` section, and it worked. 

None of this is formally documented anywhere unfortunately 😂😅. Had to scrap it together through trial and error and bits from various Stack Overflow threads. Even the [documentation](https://github.com/facebook/react-native/tree/main/packages/community-cli-plugin#bundle) for the React Native CLI was difficult to find

VAMobile/src/components/MilitaryBranchLogo.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { BranchOfService, BranchesOfServiceConstants } from 'api/types'
+import { useTheme } from 'utils/hooks'
+
+export type MilitaryBranchLogoProps = {
+  branch: BranchOfService
+  width: number
+  height: number
+}
+
+export const MilitaryBranchLogo: FC<MilitaryBranchLogoProps> = ({ branch, width, height }) => {
```
TimRoe, line 13: Only wandering into this branch out of curiosity of how it sorted out importing directly from the package... but a nitpick: highlighting that the vernacular used by the various branches (granted not uniformly) tends to be they are the "emblem" for the service--in contrast to the "seal" which is reserved for use by the department itself. Some of their brand guides are rather emphatic that their "emblem" is more distinguished than a "logo" as a representation.

theodur: Good callout Tim. Ironically, I almost named this `MilitaryBranchSeal` but decided not to for that reason. Emblem definitely makes sense, I'll probably rename this to `MilitaryBranchEmblem`

Sparowhawk (approved):

dumathane (dismissed):

TKDickson (approved):

dumathane (approved):

PR #9820, opened 2024-10-08
Automation/5721-chanel-add-flag-to-bypass-realease-check-date-release-branch

IsraelleHub:
## Description of Change
Currently the release_branch.sh script runs on a CRON job and contains a check for the to date so it doesn't run in non-release weeks. We've now added a manual trigger to that workflow so we can send out a release at any point (for hotfixes and such). With that update we need to change the release_branch.sh script so the date check can be bypassed if we're running it manually.

This will have to be tested locally since we don't want to actually trigger the workflow.

## Testing
Test was done locally. See screenshot below:
- [ ] Cron Schedule: The cron job itself will trigger the execution of the script at the scheduled time, regardless of the internal conditions.
- [ ] The workflow new_release_branch.yml has been updated to reflect the change

Conditions in the Script:
- [ ] Without the --bypass-date-check flag: The script will check if it's at the correct two-week interval using the date calculation logic. If it's not the right time for a release branch, the script will exit without making any changes.
- [ ] With the --bypass-date-check flag: If the --bypass-date-check flag is passed, the date condition will be ignored, and the script will proceed with creating the release branch, even if it's not the scheduled time for a branch.
Screenshot 2024-10-08 at 3 56 42 PM

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/.github/workflows/new_release_branch.yml
+++ b/.github/workflows/new_release_branch.yml
@@ -11,14 +11,17 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       environment_name: ${{ steps.check-environment.outputs.env_name }}
+      manual_trigger: ${{ steps.check-environment.outputs.is_manual }}
     steps:
       - name: Check environment
         id: check-environment
         run: |
           if [[ ${{ github.event_name }} == 'workflow_dispatch' ]]; then
             echo ::set-output name=env_name::'manual-release'
+            echo ::set-output name=is_manual::'true'
           else
             echo ::set-output name=env_name::''
+            echo ::set-output name=is_manual::'false'
           fi
   cut_release_branch:
     name: Cut Release Branch
@@ -41,4 +44,9 @@ jobs:
       - name: Cut release branch
         env:
           GITHUB_TOKEN:  ${{ secrets.GH_ACTIONS_PAT }}
-        run: ./release_branch.sh
+        run: |
+          if [[ "${{ needs.get-workflow-environment.outputs.manual_trigger }}" == "true" ]]; then
+            ./release_branch.sh --bypass-date-check
+          else
+            ./release_branch.sh
+          fi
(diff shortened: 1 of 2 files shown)


IsraelleHub:
@narin can you please take a look. I still need to uncomment all 'git push' on the script and also roll out the release branch to it's original.

.github/workflows/new_release_branch.yml
```diff
@@ -11,14 +11,17 @@ jobs:
...
     outputs:
       environment_name: ${{ steps.check-environment.outputs.env_name }}
+      manual_trigger: ${{ steps.check-environment.outputs.is_manual }}
     steps:
       - name: Check environment
         id: check-environment
         run: |
           if [[ ${{ github.event_name }} == 'workflow_dispatch' ]]; then
             echo ::set-output name=env_name::'manual-release'
+            echo ::set-output name=is_manual::'true'
```
narin, line 21: Do we need this additional variable? Since `env_name` is already being set, we should be able to just check if `env_name == "manual-release"` below instead of manual_trigger.

.github/workflows/new_release_branch.yml
```diff
@@ -41,4 +41,9 @@ jobs:
       - name: Cut release branch
         env:
           GITHUB_TOKEN:  ${{ secrets.GH_ACTIONS_PAT }}
-        run: ./release_branch.sh
+        run: |
+          if [[ "${{ needs.get-workflow-environment.outputs.environment_name }}" == "manual-release" ]]; then
```
narin, line 45: @timwright12 Can you double check this output var since you worked on it originally? I see that there's mixed usage between `environment_name` and `env_name` above where it's set around line 19.

narin (commented):
Waiting on some feedback from Tim W but otherwise this looks good.

.github/workflows/new_release_branch.yml
timwright12: yeah I think it's OK, just getting remapped for this job

timwright12 (approved):

PR #9821, opened 2024-10-08
feature/6975-Chika-AddAnalyticForWebviewFailEvent

cadibemma:
## Description of Change

- Added `vama_webview_fail` to capture any time a webview crashes/reports an error similar to the `vama_login_fail`. It would also capture a parameter `error` that stores the returned error message and the link for which URL the webview launched so we can differentiate between where in the app it occurred.

#### Analytics Log
image

## Testing

- [x] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


dumathane:
Lets move the WebviewError and json.stringify into the webview screen code and not the analytics definition, okay?

dumathane (changes requested):
Lets move the WebviewError and json.stringify into the webview screen code and not the analytics definition, okay?

dumathane (approved):

PR #9864, opened 2024-10-10
feature/9265-Binny-PushLoginFlowChanges

dumathane:
## Screenshots/Video

<details>
<summary>Biometric Screen Changes (Ignore BitterBoldHeading i've rectified that to the normal)</summary>
<details>
<summary>FaceID Before/After</summary>

</details>
<details>
<summary>TouchID Before/After</summary>

</details>
<details>
<summary>Face Recognition Before/After</summary>

</details>
<details>
<summary>Fingerprint Before/After</summary>

</details>
<details>
<summary>Iris Before/After</summary>

</details>
</details>
<details>
<summary>New Notification Request Screen</summary>

</details>

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/e2e/tests/AvailabilityFramework.e2e.ts
```diff
@@ -131,13 +131,13 @@ const AFNavigationForIndividual = [
...
-    // 'Claims history',
-    // 'Received December 05, 2021',
-    // 'Submit evidence',
+  // 'Claims.e2e',
+  // 'WG_SubmitEvidence',
+  // 'Benefits',
+  // 'Claims',
+  // 'Claims history',
+  // 'Received December 05, 2021',
+  // 'Submit evidence',
```
dumathane, line 140: This is the linter, not me. /shrug

VAMobile/src/screens/BiometricsPreferenceScreen/BiometricsPreferenceScreen.test.tsx
```diff
@@ -44,55 +46,74 @@ context('BiometricsPreferenceScreen', () => {
...
 
   it('initializes correctly', () => {
-    expect(screen.getByRole('header', { name: 'Do you want to allow us to use Touch ID for sign in?' })).toBeTruthy()
     expect(
-      screen.getByText(
-        'Touch ID lets you use your fingerprint to sign in to this app.\nYou can always change this later in your app settings.',
-      ),
+      screen.getByRole('header', {
+        name: t('biometricsPreference.doYouWantToAllow', {
+          biometricsText: t('biometric.touchID'),
```
dumathane, line 52: Intentionally not calling the function we use on the screen here to continue checking the function logic with the test.

VAMobile/src/store/slices/authSlice.ts
```diff
@@ -138,6 +156,10 @@ export const completeFirstTimeLogin = (): AppThunk => async (dispatch) => {
   dispatch(dispatchSetFirstLogin(false))
 }
 
+export const completeRequestNotifications = (): AppThunk => async () => {
+  await AsyncStorage.setItem(NOTIFICATION_COMPLETED_KEY, FIRST_NOTIFICATION_STORAGE_VAL)
+}
+
```
theodur, line 162: This probably doesn't need to be a function, since it's just setting a value in Asnyc Storage.

VAMobile/src/store/slices/authSlice.ts
```diff
@@ -171,6 +193,20 @@ export const checkFirstTimeLogin = (): AppThunk => async (dispatch) => {
...
+  if (IS_TEST) {
+    // In integration tests this will change the behavior and make it inconsistent across runs
+    dispatch(dispatchSetNotificationsPreferenceScreen(false))
+    return
+  }
+
+  const setNotificationsPreferenceScreenVal = await AsyncStorage.getItem(NOTIFICATION_COMPLETED_KEY)
+  console.debug(`checkRequestNotificationPreferenceScreen: is ${!setNotificationsPreferenceScreenVal}`)
+
+  const shouldShowScreen = !setNotificationsPreferenceScreenVal
```
theodur, line 206: I don't see any logic for resetting `NOTIFICATION_COMPLETED_KEY` at logout. If a previously logged-in user had notifications enabled, it looks like a new user signing in wouldn't see the RequestNotificationScreen?

VAMobile/src/store/slices/authSlice.ts
```diff
@@ -94,6 +98,8 @@ export const initialAuthState: AuthState = {
   displayBiometricsPreferenceScreen: false,
   showLaoGate: false,
   authParamsLoadingState: AuthParamsLoadingStateTypeConstants.INIT,
+  requestNotificationsPreferenceScreen: false,
```
theodur, line 101: I might be missing something, but it seems like `requestNotificationsPreferenceScreen` might be overlapping with `NOTIFICATION_COMPLETED_KEY` in async storage. Couldn't we just use `NOTIFICATION_COMPLETED_KEY` to determine whether to show the RequestNotificationScreen?

VAMobile/src/store/slices/authSlice.ts
```diff
@@ -79,6 +81,8 @@ export type AuthState = {
   authorizeStateParam?: string
   authParamsLoadingState: AuthParamsLoadingStateTypes
   successfulLogin?: boolean
+  requestNotificationsPreferenceScreen?: boolean
+  requestNotifications?: boolean
```
theodur, line 85: Since this state variable is related to notifications (and also since the authSlice will be going away soon) this should probably be moved to the [notification context](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/de30b5a69ae38c4a3dc3e68f3df98e46448e3e07/VAMobile/src/components/NotificationManager/NotificationManager.tsx#L22-L27)

theodur (dismissed):

VAMobile/src/store/slices/authSlice.ts
dumathane: Removed. I needed to drop the await in order to include it in the screen itself, but that shouldn't be a problem I don't think.

VAMobile/src/store/slices/authSlice.ts
dumathane: This is a real head scratcher for me and i'll need to bug QA about this tomorrow.

VAMobile/src/store/slices/authSlice.ts
dumathane: I will do this in the second part of this ticket. Adding an AC.

VAMobile/src/store/slices/authSlice.ts
theodur: Yeah that shouldn't cause any issues since there's not anything that needs to wait for async storage to finish setting the value

VAMobile/src/store/slices/authSlice.ts
theodur: Yeah, we could just use the value from async storage since it stays in sync with `requestNotificationsPreferenceScreen` anyway. To kill 2 birds with 1 stone, we could create a query for getting the `NOTIFICATION_COMPLETED_KEY` value, that way it'll be cached and easier to access. And when the app launches, it'll automatically load from async storage. Kinda like how we do with [useSystemNotificationsSettings](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/de30b5a69ae38c4a3dc3e68f3df98e46448e3e07/VAMobile/src/api/notifications/getSystemNotificationSettings.tsx#L11).

VAMobile/src/store/slices/authSlice.ts
dumathane: So, the fact that this tangles with OS specific permission granting and isn't actually tied to our server side notification manager is what is selling me that we should do nothing here. This is very similar to biometric preferences. When you logout of the app we do not reset your biometric preferences, we continue to allow you to use them and if another user signed on we wouldn't re-prompt. This being the 0.1% issue is another reason I don't want it to hold us up. Thoughts?

cadibemma (approved):

rbontrager (approved):

cadibemma (approved):

PR #9919, opened 2024-10-18
spike/9918-rachael-detox-add-what-to-run-to-code

rbontrager:
## Description of Change

The detox mapping file was always getting out of date/had multiple misspellings that was causing it to not run the correct files when it should.  To combat this I've reduced the numbers of files specifically checked as much as possible and added more encompassing directory checks.  I've also add jobs that will check that files/directories changed are in detox mapping file, that the directory/file names are spelled correctly, and that the e2eNames are spelled correctly.  If any of these jobs fail find_detox_tests_to_run will fail and cause the e2e_android test to fail and appear as a failure in the PR check.

## Screenshots/Video

Running test: https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/11798447379

Misspelled e2eNames
e2eNamesMisspelled

Misspelled directoryNames
DirectoryNameMisspelled

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


timwright12:
@rbontrager Since we're having spelling issues, is there a way to auto generate the mappings we need?

rbontrager:
@timwright12 Not without adding unused variables to the code (so we could see what e2e tests needed to be used).  I set up a meeting with the engineers and that was determined not to be an optimal solution.  What I did in this ticket seemed like a good compromise.  Yes you still have to manually type in stuff to the mapping file but the workflow should catch if you have mistyped the file/directory name or mistyped detox test name and error out. It also makes sure that all files changed are in the mapping in the first place and errors out if they aren't (which should make it easier to keep it up to date).  I ran my logic against everything I typed into the mapping file and fixed all the typos caught so it should be up to date (well up to date as of 3 weeks ago).

timwright12:
@rbontrager ok sounds good, can you also make sure documentation around this is up to date?

timwright12 (approved):

TKDickson (approved):

Issue #9772, opened 2024-10-03
Clean up and update docs to reflect current state of authentication via SIS


jperk51:
ACs:
- [ ] Remove references to IAM
- [ ] Create new diagrams to express how we interact with SIS
- [ ] Add info about how we are able to exchange mobile tokens for web cookie tokens
- [ ] Add anything else you think is useful for people to know about how SIS work with mobile
- [ ] Add links to SIS docs

jperk51:
@StacyB2023 we should probably make new epic similar to the QART epic for work we are doing mostly in preparation for onboarding external teams.

PR #9976, opened 2024-10-22
Remove iam from docs

Tonksthebear:
## Description of Change
This removes references to IAM from backend docs.

Ticket: https://github.com/department-of-veterans-affairs/va-mobile-app/issues/9772

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/documentation/docs/Engineering/BackEnd/Architecture/Auth Diagrams.md
```diff
@@ -14,9 +15,10 @@ sequenceDiagram
     App-->>User: Display auth result
 ```
 
+### Biometrics Flow for Username and Password Initial Login
```
Tonksthebear, line 18: I added markdown titles to each graph so that it's more obvious when there is a rendering error. I spent over an hour trying to prevent the rendering error but had no luck. From what I can tell, there is a conflict with the IDs of the mermaid graphs and there isn't a way to hardcode unique IDs to the graphs. At least with these titles on the page, you can refresh until each section is populated

jperk51: I think you might be able to solve this using aliases. https://mermaid.js.org/syntax/sequenceDiagram.html#aliases

jperk51 (changes requested):

VAMobile/documentation/docs/Engineering/BackEnd/Architecture/Auth Diagrams.md
jperk51: If not I would just screenshot the diagrams and add them as images to avoid the reloading issue.

jperk51: > It looks like aliases are for actors though, not the chart itself. Unless I'm misunderstanding the docs this wouldn't help, unless you're thinking similarly named actors are what's causing the charts to freak out?

I thought the actors could be causing the issue. Worth trying or just using images

Tonksthebear: Sadly it did not fix the issue. I updated with screenshots and left the code blocks so that we can regenerate them as needed

jperk51 (approved):
lgtm

timwright12 (approved):

Sparowhawk (approved):

PR #10039, opened 2024-10-25
9205 updating years in demo mode

### LINKED ISSUE #9205: Update Demo Mode to Align Better Align w/ Current App
**Description:** 
As VAHB, I would like to update the VAHB Demo Mode so that it better aligns with the current time and functionality so that those using demo mode knows it been updated and it works just like the app should 

**Notes:**

- Derived from ticket [7699](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7699#issuecomment-2253695551)
- Add it to QA and Release team though we don't have the ability to do this - so that we can coordinate with teams to get this work completed ; may need to split out the tickets 

**ACs:**

- Determine if we need saved data for Direct Deposit + if so add 
- Determine if we need saved data for Preferred Name + if so add 
- Mail in messages are from 2021 - can we update to make it more timely 
- Payment history is from 2016 - can we update to make it more timely

TKDickson:
## Description of Change

Closes #9205 

 In demo mode:
* Payment history updated to 2023 and 2024 (instead of 2016 and 2017)
* Messages updated to 2024

Also updated detox tests to account for both of the year changes above. (I may need to update unit tests - although I'm certainly hoping not - but we'll find out when I change this PR from draft to real.)

## Testing

Tested detox for [iOS](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/11562088470) and [Android](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/11562798987) (messages and payments, although my links here will be messages-only since the payments ones also run against the PR itself)
* The Android AF Use Case 2 with update is also failing in develop, so not addressing here. Wrote up ticket #10058 

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


Sparowhawk:
@TKDickson I know we have it somewhere, but maybe this should be updated to use the "x days from current date" function that @alexandec created that we use in some unit tests already and for Appointments.

alexandec:
Good call Dylan, the feature I added parses all the demo mode mock JSON files looking for patterns like `{{now + 36 days}}` and replacing them with the corresponding date. Here's the documentation:

```
 * Replace double curly brace date expression in mock file with ISO date.
 * For example, {{now + 5 days}} is replaced with 2023-08-08T10:58:58.003-06:00
 * Units can be days, weeks, months, or years, and you can add or subtract.
 * You can add an optional format at the end. Options:
 *   short: month-day-year (05-30-2023)
 *   shortReversed: year-month-day (2023-07-11)
 *   year: year only (2023)
 *   utc: timestamp in UTC time (2023-08-10T22:04:16.695Z)
 *   (default): timestamp in local timezone (2023-11-08T16:04:16.693-07:00)
```

Oh, and you can use minutes as well like `{{now + 45 minutes}}`. The replacement happens when demo mode is turned on, so that's what "now" is considered to be for this purpose.

TKDickson:
@Sparowhawk @alexandec appreciate the info, and wrote up #10059 for that idea (linked it to the comment above). That'd be great to do at some point - and I'm going to consider it a follow-up, since it'll involve digging into how/if we'll be able to account for relative IDs as part of a testID in detox, and I'm not volunteering to tack that work onto this PR 😄

alexandec:
@TKDickson makes sense as a followup, and thanks for creating that ticket. A similar approach to what we use in Appointments.e2e.ts might work, where the E2E tests generate date strings that match the demo mode date strings. Like:

```
const fiveMonthsEarlier = todaysDate.minus({ months: 5 }).startOf('month').startOf('day')
const threeMonthsEarlier = todaysDate.minus({ months: 3 })

 await element(
      by.text(
        fiveMonthsEarlier.monthShort +
          ' ' +
          fiveMonthsEarlier.year +
          ' - ' +
          threeMonthsEarlier.monthShort +
          ' ' +
          threeMonthsEarlier.year,
      ),
    ).tap()
```

rbontrager (approved):

timwright12 (approved):

PR #10040, opened 2024-10-25
bug/9615-BiometricLoginRefreshTokenFix

Sparowhawk:
## Description of Change
Set the storage to AES for the keychain token. AES and FB according to this thread 
https://github.com/oblador/react-native-keychain/issues/262

Does not throw this error Wrapped error: javax.crypto.IllegalBlockSizeException which seems to be the bulk of the spike

Biometric when not specified seems to pick one based on the library settings, which I haven't been able to determine what the default is, but is likely RSA or KC which both cause this issue

## Screenshots/Video
N/A

## Testing
Yarn test

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations
Biometrics should work and not kick you back out of the app and force you to log back in to use biometrics

## PR Checklist

## For QA


dumathane (changes requested):
tests still failing.

Sparowhawk:
@dumathane updated unit tests

dumathane (approved):

PR #10041, opened 2024-10-25
 CU/9882-rachael-detox-add-comments-to-appeals-appealsExpanded

rbontrager:
## Description of Change

Add comments at the top of the appeals and appealsExpanded detox tests.

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


TKDickson:
I like this format generally, and don't think appeals needs anything more. I'd like a comment for "what the heck is happening with appeals expanded starting at line 163". I assume it's some sort of error handling along the lines of "if one appeal in the chain gets messed up, we can still check later ones" but would love a comment to assist with how/with what/when I'd update it.

TKDickson (approved):

PR #10101, opened 2024-11-04
Update e2e_android.yml

rbontrager:
## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -82,7 +82,6 @@ jobs:
       message: 'Starting E2E Android tests. Please see :thread: for results. This process may take a while.'
 
   find_detox_tests_to_run:
-    if: github.event_name == 'pull_request' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     uses: ./.github/workflows/e2e_detox_mapping.yml
   
   output_detox_tests_to_run:


theodur (commented):
Do we also need to update the logic here?
https://github.com/department-of-veterans-affairs/va-mobile-app/blob/53917c177c8dec3b3010b081cac9b5474c4f4278/.github/workflows/e2e_android.yml#L101-L102

Since only `package.json` and `yarn.lock` are typically the only changed files in Dependabot PRs, wouldn't `output_detox_tests_to_run` see an empty array?

.github/workflows/e2e_android.yml
```diff
@@ -99,8 +98,8 @@ jobs:
       - name: 'Get Matrix Value'
         id: matrix_value
         run: |
-          if [[ "${{ inputs.run_full_test }}" == "true" ]] || [[ ${{ github.event_name }} == 'schedule' ]] || [[ "${{ inputs.tests_to_run}}" != "" ]]; then
-            if [[ "${{ inputs.run_full_test }}" == "true" ]] || [[ ${{ github.event_name }} == 'schedule' ]]; then
+          if [[ "${{ inputs.run_full_test }}" == "true" ]] || [[ ${{ github.event_name }} == 'schedule' ]] || [[ "${{ inputs.tests_to_run}}" != "" ]] || [[ ${{ github.event_name }} == 'push' ]]; then
```
theodur, line 101: Wouldn't this run all tests on branches anytime a push happens? If we only want to change the logic for just Dependabot PRs we could do something like

```yml
[[ "${{ github.event.pull_request.user.login }}" == "dependabot[bot]" ]]
```

IsraelleHub (approved):

PR #10214, opened 2024-11-19
CU/9943-rachael-detox-add-comments-to-navigation

rbontrager:
## Description of Change

Add comments to navigation

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/e2e/tests/Navigation.e2e.ts
+++ b/VAMobile/e2e/tests/Navigation.e2e.ts
@@ -1,3 +1,9 @@
+/*
+Description:
+Detox script that runs dark mode, landscape, and font size accesability tests
+When to update:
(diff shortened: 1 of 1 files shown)


VAMobile/e2e/tests/Navigation.e2e.ts
```diff
@@ -1,3 +1,9 @@
+/*
+Description:
+Detox script that runs dark mode, landscape, and font size accessibility tests
```
TKDickson, line 3: Does it make sense to add a line describing the difference between running this in the context of the full suite, vs a subset? Or would you expect folks to step through the code to make that leap?

I'd personally add it, since I've fielded that question from our FE before.

TKDickson (approved):

PR #10219, opened 2024-11-19
Cu/10086 tkd detox utils constants

### LINKED ISSUE #10086: CU - utils detox: replace strings with constants, if any remaining
Split up the utils cleanup work into three tickets:

#9851: for things with existing IDs that we just aren't using in the utils file, replace by.text with by.id. And, if those are being defined in 'downstream' scripts, move those to this one (aka move the carousel skip button from Onboarding to utils, since we need it in utils)
#10085: creating net-new IDs for things we can create IDs for, and replacing by.text with by.id for those
THIS ticket: replacing any calls that are using strings ex by.text('Remote Config') to use a constant instead, regardless if they're text or id

TKDickson:
## Description of Change

Closes #10086 
- In the detox utils file, replacing all of the strings still directly used in matchers, with constants
- In cases where that particular constant was already defined in a downstream file, replaced it with the new one from utils
- In cases where downstream scripts were also referencing one of these strings directly, also replaced with new constant from utils

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/e2e/tests/utils.ts
```diff
@@ -458,21 +473,21 @@ export async function enableAF(AFFeature, AFUseCase, AFAppUpdate = false) {
...
   }
-  await element(by.id('AFTypeTestID')).replaceText(AFUseCase)
-  await element(by.id('AFTypeTestID')).tapReturnKey()
-  await element(by.id('AFErrorMsgTitleTestID')).replaceText('AF Heading Test')
-  await element(by.id('AFErrorMsgTitleTestID')).tapReturnKey()
-  await element(by.id('AFErrorMsgBodyTestID')).replaceText('AF Body Test')
-  await element(by.id('AFErrorMsgBodyTestID')).tapReturnKey()
+  await element(by.id(CommonE2eIdConstants.AF_TYPE_INPUT_ID)).replaceText(AFUseCase)
+  await element(by.id(CommonE2eIdConstants.AF_TYPE_INPUT_ID)).tapReturnKey()
+  await element(by.id(CommonE2eIdConstants.AF_ERROR_MSG_TITLE_INPUT_ID)).replaceText('AF Heading Test')
```
rbontrager, line 478: I think you can change the value of 'AF Heading Test' to CommonE2eIdConstants.AF_ERROR_MSG_TITLE_ENTERED_TEXT and can probably do the same to 'AF Body Test'.

rbontrager (approved):

PR #10298, opened 2024-12-03
Chore/9630-UseDesignSystemSnackbarComponent

oddballdave:
## Description of Change

- Completely removed the old global SnackBar implementation and replaced it with the design system Snackbar component
- Incorporated the use of the Snackbar provider and hook from the design system
- Patched the design system Snackbar to workaround dependency issues with the useEffect hook
- Implemented automatic dismissal of the Snackbar when the user navigates away from the current screen
- Removed SnackbarMessages and switched to using translated text directly
- Modified unit tests to work correctly with the new Snackbar implementation
- Cleaned up unused files, constants, string resources and related portions of the app-level theme

## Screenshots/Video

[Android (Before)](https://github.com/user-attachments/assets/547febdf-4bc5-4a89-98a9-74e183af19ba)
[Android (After)](https://github.com/user-attachments/assets/c34ac63a-c35d-4318-9b2a-f9068a866596)
[iOS (Before)](https://github.com/user-attachments/assets/f81adefd-4fbe-4af4-a0ed-2c292c50ac72)
[iOS (After)](https://github.com/user-attachments/assets/a9a47421-79b1-403a-9add-5fdcd3ac2e5c)

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


Sparowhawk (dismissed):

VAMobile/src/constants/analytics.ts
```diff
@@ -1149,14 +1149,14 @@ export const Events = {
...
-    }
-  },
+  // vama_snackbar_null: (location: string): Event => {
+  //   return {
+  //     name: 'vama_snackbar_null',
+  //     params: {
+  //       location,
+  //     },
+  //   }
+  // },
```
TimRoe, line 1159: Snooping out of curiosity from the design system side:
```suggestion
```
not sure there's value in retaining a commented out analytics event--one that in theory is hopefully eliminated with the new Snackbar since it was only added to track down a weird and rare crash where somehow the global Snackbar was not being initialized.

oddballdave: I think I saw a few other events commented out and assumed this was done to keep the history in tact. If that's not necessary, then it can obviously be removed.

timwright12:
@oddballdave please make sure to not merge until QA has tested and approved. This is a large change, so it will take some time.

narin (changes requested):
Looks great for the most part. We'll consider updating the component library to use the patch you created.

Reiterating what we discussed on the call, for Snackbars that need to appear on screens using templates without the nav bar showing such as `<FullScreenSubtask />` or `<LargePanel />`, if a Snackbar is shown and the user isn't immediately navigated back to a screen that has the nav bar showing, the `offset` option should be included with `snackbar.show()` to account for the lack of nav bar.  This probably would occur mostly with error message Snackbars.  You can use `theme.dimensions.snackBarBottomOffset` as the offset which automatically calculates the correct offset based on OS.

i.e. 

```tsx
snackbar.show(t('secureMessaging.startNewMessage.sent.error')'), { offset: theme.dimensions.snackBarBottomOffset })
```

VAMobile/src/constants/analytics.ts
oddballdave: This event has been removed.

oddballdave:
All known issues resolved and ready for a re-review. For reference, Jessica Wooden provided this guidance for the Snackbar design.

> Here's a visual guide for where the Snackbar should appear on the screen if the nav bar is present or not. The spacing to the left/right and below the snackbar is set to 20.

image

VAMobile/patches/@department-of-veterans-affairs+mobile-component-library+0.27.1.patch
```diff
@@ -0,0 +1,67 @@
+diff --git a/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/Snackbar.tsx b/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/Snackbar.tsx
+index 6b2f54d..8931ba9 100644
+--- a/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/Snackbar.tsx
++++ b/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/Snackbar.tsx
+@@ -160,6 +160,7 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
+       minHeight: 44,
+       padding: spacing.vadsSpaceSm,
+       marginHorizontal: spacing.vadsSpaceLg,
++      marginBottom: spacing.vadsSpaceLg,
```
TimRoe, line 9: I see the justification for this in another comment:
> I made another patch to the Snackbar component to update the navigation bar height (more accurate) and to add a bottom margin to the Snackbar. This seemed like the right place to fix things since we are already accounting for the left/right margins and navigation bar height in the component itself versus relying on the app to provide these properties.

The reason bottom margin was not included in the component is two fold:
1. It's controllable via the offset to whatever an app could want so we just gave a default as apps would need to override it anyway for any special cases--if 20 pixel spacing is desired without a Nav bar, offset can be set to 20
2. Margin makes elements underneath the Snackbar (e.g. a button sticking out partially) no longer pressable within the margined space despite being visible where setting the offset via the library used by the component allows presses outside the bounds of the Snackbar to still occur as desired
   - For full disclosure: when building it, I tried to use margin instead of the library offset functionality since it was simpler, but found it resulted in the bug where you could position it above the Nav bar visually, but none of the Nav bar buttons would be able to be pressed; using the offset functionality, the Nav bar can still be interacted with as meets user expectations

The side margins were deemed acceptable to block presses as, paired with general padding for content, anything peeking should be very minimal.

I technically defer to the app team since we on the design system team long term are not responsible for how apps implement our components, but I'd recommend against patching this in as patches are frail and this is liable to break from future library changes we make to the component when updating while leveraging the offset should not.

VAMobile/patches/@department-of-veterans-affairs+mobile-component-library+0.27.1.patch
```diff
@@ -0,0 +1,67 @@
...
+ }
+diff --git a/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/useSnackbarDefaultOffset.tsx b/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/useSnackbarDefaultOffset.tsx
+index e7679fa..135c5fa 100644
+--- a/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/useSnackbarDefaultOffset.tsx
++++ b/node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/useSnackbarDefaultOffset.tsx
+@@ -1,6 +1,6 @@
+ import { useSafeAreaInsets } from 'react-native-safe-area-context'
+ 
+-const NAV_BAR_HEIGHT = 60
++const NAV_BAR_HEIGHT = 56
```
TimRoe, line 64: This is somewhat a design decision, but am not sure pixel perfection is necessary here.

Similar to the above of deferring to the app team, but I'd recommend against patching it vs having 4 pixels more in what is already somewhat arbitrary padding for a "floating" element.

VAMobile/src/screens/PaymentsScreen/DirectDepositScreen/EditDirectDepositScreen/EditDirectDepositScreen.tsx
```diff
@@ -122,6 +122,7 @@ function EditDirectDepositScreen({ navigation, route }: EditDirectDepositProps)
           if (!routingNumberError) {
             snackbar.show(t('directDeposit.saved.error'), {
               isError: true,
+              offset: 0,
```
TimRoe, line 125: I think Narin had already suggested it somewhere, but also suggesting having the offset for non-Nav screens set in the app theming instead of having so many hard coded 0's. Make it much easier to change in the future if it's desired vs updating ~25 spots.

VAMobile/patches/@department-of-veterans-affairs+mobile-component-library+0.27.1.patch
oddballdave: Thanks for the feedback. Just so we're all on the same page, the left screenshot below is the default location where the Snackbar gets positioned when a bottom navigation bar is present and no offset is specified. There are 4 units of vertical spacing between the top of the navbar and the Snackbar. The right screenshot shows the changes with the patch applied (20 units of vertical spacing). If we can live with the spacing on the left screenshot, then I can remove the patch easily enough. If we want more spacing but no patch, then the app will have to start passing offsets into all Snackbars. Please advise.

<div style="display: flex; justify-content: space-around;">
  
  
</div>

VAMobile/src/screens/PaymentsScreen/DirectDepositScreen/EditDirectDepositScreen/EditDirectDepositScreen.tsx
oddballdave: Agree, I will make this change.

VAMobile/patches/@department-of-veterans-affairs+mobile-component-library+0.27.1.patch
TimRoe: I think the 4 units spacing is fine, but ultimately seems like a design decision. I believe the position is what it was with the old Snackbar so it'd be maintaining existing functionality and didn't come up during creating it or we'd likely have gone with a different default. As a "floating" element it can appear directly above other content regardless if it's 4 or 20 pixels from the Nav bar--the crux is not blocking the Nav bar so it's usable.

If the increased spacing is determined the way: I think a patch is fair way to go about it, but it would be better to update the default offset to 76 instead of using `marginBottom`. Longer term, we have done research into global overrides for certain design system parameters and the Snackbar offset definitely would be a good use case--but, in the absence of that yet existing, patching in a different default value seems like a reasonable workaround since when the global override exists then the only code change would be removing the patch and setting the global override in one spot.

Not sure if @Sparowhawk or anyone else on the app side had thoughts since ultimately from the design system side we're just providing recommendations based on how the components were built/designed and consumer apps can do what they want.

TimRoe: As a circle back: [created a ticket](https://github.com/department-of-veterans-affairs/va-mobile-library/issues/609) to add a global override for the default Snackbar offset so it's documented as a desired future enhancement.

TimRoe (dismissed):
Approving for my part (still needs app team approval) based on the changes.

TimRoe (approved):

oddballdave:
Moved back to draft until this [DS library PR](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/617) can be incorporated.

Eallan919:
These PR's are blocked because they depend on updating the DS Library with the previously mentioned PR's and they also need to have a final review by an authorized user: [#10298](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/10298) & [#10392](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/10392).

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/AskForClaimDecision/AskForClaimDecision.tsx
```diff
@@ -102,9 +102,14 @@ function AskForClaimDecision({ navigation, route }: AskForClaimDecisionProps) {
     const mutateOptions = {
       onSuccess: () => {
         setSubmittedDecision(true)
-        showSnackBar('Request sent', dispatch, undefined, true, false, true)
+        snackbar.show('Request sent')
       },
-      onError: () => showSnackBar('Request could not be sent', dispatch, () => onSubmit, false, true),
+      onError: () =>
+        snackbar.show('Request could not be sent', {
```
matt-guest-wilcore, line 108: Should we localize 'Request could not be sent'? I see all the other snackbar messages are using localized strings.

VAMobile/src/screens/BenefitsScreen/ClaimsScreen/ClaimDetailsScreen/ClaimStatus/ClaimFileUpload/AskForClaimDecision/AskForClaimDecision.tsx
```diff
@@ -102,9 +102,14 @@ function AskForClaimDecision({ navigation, route }: AskForClaimDecisionProps) {
     const mutateOptions = {
       onSuccess: () => {
         setSubmittedDecision(true)
-        showSnackBar('Request sent', dispatch, undefined, true, false, true)
+        snackbar.show('Request sent')
```
matt-guest-wilcore, line 105: Same as the comment below

oddballdave: Done

matt-guest-wilcore:
Adding comment here since slack is still down but I think I got every Snackbar to show in the app and I didn't see any issues. Code-wise everything looks good too, I had a couple comments but I see you already made the updates. Overall, nice work!

Kim said he was looking into this as well so let's see if he has any feedback.

kimmccaskill-oddball:
I did some BRIEF testing with VoiceOver on a physical device and it seems good to go from that front

VAMobile/src/screens/HealthScreen/SecureMessaging/CancelConfirmations/ComposeCancelConfirmation.tsx
```diff
@@ -70,7 +66,7 @@ export function useComposeCancelConfirmation(): [
           const params: SaveDraftParameters = { messageData: message, replyID: replyToID, messageID: draftMessageID }
           const mutateOptions = {
             onSuccess: () => {
-              showSnackBar(snackbarMessages.successMsg, dispatch, undefined, true, false, true)
+              snackbar.show(t('secureMessaging.draft.saved'))
```
kimmccaskill-oddball, line 69: Just a quick question, but is there any functionality we lose with some of these params in `showSnackBar` like the `isUndo`?

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/EditAddressScreen/EditAddressScreen.tsx
```diff
@@ -220,15 +211,13 @@ function EditAddressScreen({ navigation, route }: IEditAddressScreen) {
...
-          removalSnackbarMessages.errorMsg,
-          dispatch,
-          () => deleteAddress(currentAddressData, mutateOptions),
-          false,
-          true,
-        ),
+        snackbar.show(t('contactInformation.residentialAddress.removed.error'), {
+          isError: true,
+          offset: theme.dimensions.snackBarBottomOffset,
+          onActionPressed: () => deleteAddress,
```
kimmccaskill-oddball, line 219: Just based on other examples of the replacement, should this be `() => deleteAddress(currentAddressData, mutateOptions)` ?

VAMobile/src/screens/HealthScreen/SecureMessaging/CancelConfirmations/ComposeCancelConfirmation.tsx
oddballdave: Researched this a bit. Even though the undo param is true in many cases, the actionedPressed param is always undefined which means the existing snackbar implmentation just shows a dismiss button. So essentially there are no undo operations implemented in the entire app and we do not lose functionality.

VAMobile/src/screens/HomeScreen/ProfileScreen/ContactInformationScreen/EditAddressScreen/EditAddressScreen.tsx
oddballdave: Yes, great catch! Fix committed.

VAMobile/src/screens/HealthScreen/SecureMessaging/CancelConfirmations/ComposeCancelConfirmation.tsx
oddballdave: Actually, the ViewMessageScreen.tsx seems to be the one screen that does have undo functionality when you move a message from one folder to another. I have reviewed the logic and it looks correct, but would you mind retesting this to make sure the undo works correctly?

kimmccaskill-oddball (approved):

PR #10362, opened 2024-12-10
chore/5057-streamlineUseIsScreenReaderEnabledHook

alexandec:
## Description of Change

Switch to the useIsScreenReaderEnabled hook in the component library. Remove our version of the hook.

## Screenshots/Video

Toggling VoiceOver:

https://github.com/user-attachments/assets/912ef757-1998-4425-932f-45e148a3d9ea

Console output when toggling, showing events:

```
setAnalyticsUserProperty vama_uses_screen_reader true
setAnalyticsUserProperty vama_uses_screen_reader false
setAnalyticsUserProperty vama_uses_screen_reader true
```

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

Check usage of useIsScreenReaderEnabled for regressions:

- CategoryLandingTemplate
- FeatureLandingAndChildTemplate
- HeaderBanner
- CollapsibleMessage
- useAutoScrollToElement (AlertWithHaptics, ClaimPhase)
- useAccessibilityFocus (BackButton, SnackBar, HeaderBanner, GenericOnboarding)

vama_uses_screen_reader event is also good to check, but I don't think QA has a good way to check it. In the console I can see that it's being sent every time the screen reader is turned off/on. I can also see that it's sent on app load.
 
## PR Checklist

## For QA


alexandec:
@TimRoe since you worked on this in the past, if you could take a look, that'd be great

TimRoe:
@alexandec I can, but seeing this will also mention that we ended up adding this hook to the design system and it is being exported as `useIsScreenReaderEnabled` so it could potentially be simplified further to relying on that and removing the hook from the app entirely.

TimRoe (approved):
Code-wise, looks good to me.

Only critique is the PR write-up not including that testing should also validate the `useAccessibilityFocus` use cases as another hook leveraging the existing one besides the `useAutoScrollToElement` mentioned.

theodur (approved):

rbontrager (approved):

PR #10392, opened 2024-12-17
chore/8936-UseDesignSystemLoadingComponent

oddballdave:
## Description of Change
Incorporated the Design System LoadingIndicator component into the app. This is a partial replacement for the existing app-supplied LoadingComponent which handled these scenarios:

1. Showing a spinner only with a custom color.
2. Showing an inline spinner with a customized color and size.
3. Showing a spinner with text below it.

This work only covers scenario 3 because the Design System LoadingIndicator does not provide a way to customize it for the other two cases. Furthermore, to center the LoadingIndicator vertically on the screen it had to be wrapped by the existing app-supplied LoadingComponent.

After implementing these changes, all detox/e2e screen tests that showed the loading component started to fail. Research and experimentation showed that the animation logic in the component was flawed, so this was adjusted to a cleaner/safer implementation so the tests would function normally.

## Screenshots/Video
[Android (Before)](https://github.com/user-attachments/assets/e7b4ebe5-612b-47ba-b56b-064f6f30ed11)
[iOS (Before)](https://github.com/user-attachments/assets/26b900ad-60a7-4533-82ff-dea24b54ff21)
[Android (After)](https://github.com/user-attachments/assets/c757d577-a4c1-427e-84d6-69960fe5c4b0)
[iOS (After)](https://github.com/user-attachments/assets/fd20e806-e7a7-4165-9c28-e9a0694e3b06)

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


oddballdave:
Moving back to draft until [this DS library issue](https://github.com/department-of-veterans-affairs/va-mobile-library/issues/618) is fixed.

Eallan919:
These PR's are blocked because they depend on updating the DS Library with the previously mentioned PR's and they also need to have a final review by an authorized user: [#10298](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/10298) & [#10392](https://github.com/department-of-veterans-affairs/va-mobile-app/pull/10392).

matt-guest-wilcore (approved):
I tested on both iOS and Android and everything looks good. Approved!

Issue #10354, opened 2024-12-09
BUG - Sev-3 - In-App Feedback: Phone number, SSN, Email are not properly recognized if they end in punctuation


rbontrager:
## What happened?

Currently in app feedback does not properly x out a phone number, SSN, or email if the items end in punctuation.  This could cause problems if PII information is entered that ends in punctuation.

## Specs:

- Device:
- OS:
- User Account:
- Accessibility State:

## Steps to Reproduce

1. Trigger in-app feedback
2. Type in 123-12-1234 and tap 'submit feedback'. Notice that a 'Submit your feedback with personal information' dialog is displayed
3. Tap 'Edit Feedback' and add a '.' after 123-12-1234. 
4. Tap 'Submit feedback'. Notice that no dialog is displayed

## Bug Severity - BE SURE TO ADD THE SEVERITY LABEL
 
See [Bug Tracking](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/QA#issue-severity) for details on severity levels

- Impact: High Low 
- Frequency: High Low 

- 3 - Low
- 2 - High
- 1 - Critical

## Screen shot(s) and additional information

<details>
  <summary>Full JSON response for services related to issue (expand/collapse)</summary>
  

</details>

## Ticket Checklist
- [ ] Steps to reproduce are defined
- [ ] Desired behavior is added
- [ ] Labels added (front-end, back-end, global, Health, relevant feature, accessibility, etc)
- [ ] Estimate of 1 added (for front-end tickets)
- [ ] Added either to the relevant feature epic (if found during new feature implementation), or the relevant team's bug epic ([Global](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8302), [Health & Benefits](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8303), [API](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8304), [QART](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8305))

jennb33:
12/19/2024 - Possible new issue for MFS?

[...]

PR #10395, opened 2024-12-20
Bug/10354 - In-App Feedback: Phone number, SSN, Email are not properly recognized if they end in punctuation

matt-guest-wilcore:
## Description of Change

[Ticket](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/10354)

The in-app feedback screen doesn't allow for personal information such as phone numbers, SSN and emails, if they are detected then we must show the user a warning. 

As described in the ticket, if someone input an email but ended it with punctuation then we wouldn't show the user the warning. 

This PR fixes that issue, we can now detect sensitive information even if it has punctuation.

## Testing

Navigate to the developer screen to find the in-app feedback screen

- [x] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

Verify that sensitive information like emails, phone numbers and SSNs that end in punctuation trigger a warning alert.

EX:
- 123-1234-123.
- 123-12-1234?
- test@gmail.com;

## PR Checklist

## For QA


DonMcCaugheyUSDS (commented):
The `checkStringForPII()` function likely still has the inverse problem when there is _leading_ punctuation (e.g. `My SS:123-45-6789` or even `call me415-321-1234`).  Looking at `EMAIL_REGEX_EXP` and its mates, they're anchored to the start (`^`) and end (`$`).  Did you consider alternate regex patterns that might find the forbidden patterns without requiring the string to be split on whitespace?

Also, this seems like an easy function to unit test, but I didn't see any tests for it.  If I didn't miss something, would you add unit tests?

DonMcCaugheyUSDS (approved):
Approved, thanks Matt for improving this and adding tests!  Please fix the linting error and merge as it.

I am requesting some follow-on work.  You added tests for the `InAppFeedbackScreen` component, which is a correct and important to test, but you use this component to drive all the test cases for `checkStringForPII()`.

I strongly prefer that tests focus on the smallest unit of code possible, which is the `checkStringForPII()` function in this case.  Testing `InAppFeedbackScreen` is important too, but is more of an integration test than a unit test.

Small, focused tests are easier to understand and maintain and usually run significantly faster.  In this case, `InAppFeedbackScreen.test.tsx` is 177 lines long, requires loading heavyweight React libraries and creating a new `InAppFeedbackScreen` component instance for each test case and waiting asynchronously for the component to respond to events.  I bet that unit tests for just `checkStringForPII()` run 10 to 100 times faster and significantly smaller code than `InAppFeedbackScreen.test.tsx`.  And from personal experience, focused tests will make extending `checkStringForPII()` much easier (and I can envision us discovering new PII patterns to filter for).

Please keep a couple representative test cases for `InAppFeedbackScreen` and look into adding unit tests for `checkStringForPII()`.  Let me know if there's anything challenging about writing smaller, focused tests like this.

matt-guest-wilcore:
@DonMcCaugheyUSDS I was making the updates for the follow on work you mentioned and I found an issue with the `checkStringForPII` function itself. Basically, my current implementation erases the entire string and returns the censored portion but the entire string needs to be maintained with the appropriate pattern censored. This became apparent as I was writing the unit tests for `checkStringForPII`.

The fix shouldn't be too drastic but I'd wait on merging this until I get that fix in. My apologies for the delay.

kimmccaskill-oddball (approved):
Looks to address Don's concerns and I appreciate all the test cases!

Issue #7508, opened 2023-12-11
BUG - sev-3 - All - Improve UI for when elements missing in Vaccine list & details


TKDickson:
## What happened?

The upstream data has changed -- when we originally built vaccines, we built it with the assumption that a trio of fields (date, groupName, and type and dosage) would always have data. Now, at least groupName (and potentially more) can come back empty.

We've stopped the entire activity from erroring out when this occurs (#7335) but should also loop back to improve the UI for veterans who experience this issue.

[Screenshots of what this looks like in the app, from this comment in ticket 7335.](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7335#issuecomment-1850860240)

<img width="369" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBODU2QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b3f2effb47a9531ee9fa5c34f6d33ccc4cd9e760/image.png" alt="image.png" />

<img width="369" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBODk2QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--76d42225079cfaf044142406abf83c6bb7864dc7/image.png" alt="image.png" />

## Specs:

- Device:
- OS:
- User Account:
- Accessibility State:

## Desired behavior

TBD need input from UX - what should the activity look like, if we don't have data for one/all of those fields?

## Bug Severity - BE SURE TO ADD THE SEVERITY LABEL
 
See [Bug Tracking](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/QA#issue-severity) for details on severity levels

- Impact: Low 
- Frequency: Low 

- 3 - Low
- 
## Screen shot(s) and additional information

<details>
  <summary>Full JSON response for services related to issue (expand/collapse)</summary>
  

</details>

## Ticket Checklist
- [ ] Steps to reproduce are defined
- [ ] Desired behavior is added
- [ ] Labels added (front-end, back-end, relevant feature, accessibility, etc)
- [ ] Estimate of 1 added (for front-end tickets)
[...]

Issue #7335, opened 2023-11-20
BUG - sev-2 - All - Vaccines: groupname null causes vaccines page to error out to user without throwing an error


TKDickson:
## What happened?

+119, one of our standard vaccine staging test users, is now displaying an error in vaccines instead of information, because some of their vaccines have null for groupname (which is what the row in teh vaccine list is labelled with).

Looks like ([just guessing from the mock test data](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/8507d435d1c2df4d85f967cc6b4c21c96a8f2eb0/VAMobile/src/store/slices/vaccineSlice.test.ts#L71)) that we're expecting an empty string, rather than null. So maybe a backend system change that breaks our assumptions?

But the API is returning real data, so 1) we shouldn't be showing an error and 2) this error is happening "invisibly" (not showing up in analytics) so we have no idea how frequently it's happening in the wild.

## Specs:

- Device:
- OS:
- User Account: +119
- Accessibility State:

## Steps to Reproduce

See relevant JSON in the collapsible Charles section

## Desired behavior

Should have a fallback for the vaccine row label, I'm guessing?

Also we should probably figure out how to log an error for something like this in the future :sigh:

## Bug Severity - BE SURE TO ADD THE SEVERITY LABEL
 
See [Bug Tracking](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/QA#issue-severity) for details on severity levels

- Impact: High 
- Frequency: Low 

- 2 - High

## Screen shot(s) and additional information

<details>
  <summary>Full JSON response for services related to issue (expand/collapse)</summary>
  

HTTP/1.1 200 OK
Date: Fri, 17 Nov 2023 21:57:26 GMT
Content-Type: application/json; charset=utf-8
Etag: W/"298c5d2a3b590af365fbf50999168cca"
Referrer-Policy: strict-origin-when-cross-origin
Vary: Accept, Origin
X-Content-Type-Options: nosniff
X-Custom-Response-Header: eks-staging
X-Download-Options: noopen
X-Frame-Options: SAMEORIGIN
X-Git-Sha: d381ead70a7d9894bc6f738b89402583cf383604
X-Github-Repository: https://github.com/department-of-veterans-affairs/vets-api
X-Permitted-Cross-Domain-Policies: none
X-Request-Id: 80f2a314-0e4f-48b6-8fec-ea088595dd28
X-Runtime: 1.191146
[...]

PR #10398, opened 2025-01-06
Update vaccine list and details ui to handle missing vaccine name and…

kimmccaskill-oddball:
## Description of Change

[Issue #7508](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7508)
For context, initially, the bug was that the app was breaking if the vaccine name was null. That was fixed quickly [here](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/7335) without consideration for a UI fix.

With this [Figma design]([here's the Figma](https://www.figma.com/design/P1MgY68Y7Jju5PCfaJHQX4/Misc.-MFS-Stuff?node-id=58-13801&t=rSdRQkBmvtTR7qcQ-1)), on the list screen, " vaccine" was fixed to "Vaccine" and the blank date now says "No date noted". On the details screen, we've added an info alert that tells the user that we're missing some details about the vaccine and the same placeholders are filled in.

### NOTE 
I wasn't able to reproduce this with an actual user(the former issue noted user +119 which I could not replicate still) so I only tested with hard coded data

### Vaccine List Screen
Before
Screenshot 2025-01-06 at 1 19 17 PM

After
Screenshot 2025-01-06 at 1 17 37 PM

### Vaccine Details Screen
Before
Screenshot 2025-01-06 at 1 20 37 PM
After
Screenshot 2025-01-06 at 1 21 02 PM

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/screens/HealthScreen/Vaccines/VaccineDetails/VaccineDetailsScreen.tsx
```diff
@@ -77,89 +76,98 @@ function VaccineDetailsScreen({ route, navigation }: VaccineDetailsScreenProps)
       {detailsLoading ? (
         <LoadingComponent text={t('vaccines.details.loading')} />
       ) : (
-        <Box mb={contentMarginBottom}>
```
kimmccaskill-oddball, line 80: This will be easier to read if you tick `Hide whitespace` in the diff settings

DonMcCaugheyUSDS (approved):
Good work.  Thank you!

PR #10400, opened 2025-01-10
Task/86 - Incorporate Icon Component in Flagship App

matt-guest-wilcore:
## Description of Change

Updated every instance of `<VAIcon />` in the app to the new `<Icon />` component from the mobile-component-library dependancy. Doing so also required finding and updating the names of the assets to their new equivalent (ex: RightArrowInCircle is now ArrowCircleRight). 

I also had to update a few tests after adding the Icon component to some screens. For the components `BaseListItem` and `VASelector` I had to incorporate the new vads colors in order to get icons to show properly (Some didn't mesh well with the previous theme system).

I made an effort to check every screen in the app (in light and dark mode) to make sure the new icons are showing properly. I'd recommend also going through the app yourself in case I missed anything. I'm going to try to get the UX designers to have a look at this and get their input as well.

## Testing

- [x] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


kimmccaskill-oddball:
Screenshot 2025-02-11 at 3 06 08 PM
Screenshot 2025-02-11 at 3 06 48 PM
While everything seems to be working, I just want to make sure the differences are intentional. Thanks!

matt-guest-wilcore:
Yes, some icons will look different. The new Icon library sometimes doesn't have the same icons as the previous so I had to follow the Figma for guidance. Here's an [example](https://www.figma.com/design/Zzt8z60hCtdEzXx2GFWghH/%F0%9F%93%90-Component-Library---Design-System---VA-Mobile?node-id=235-674&t=cAE23rAtl3d0xNuN-0) showing the icons used for links in the app using the new library. Because the VA.gov link opens up an external web page the icon was changed to reflect that.

I'm thinking I should get UX to take a look at this just to verify everything is updated as it should. Thanks for bringing this up.

kimmccaskill-oddball (dismissed):
Gonna go ahead and approve the code pending design approval ✅

nhuckleberry6:
@matt-guest-wilcore @kimmccaskill-oddball thanks for finding those and calling them out! Good catch on removing the icon next to "Only use messages for non-urgent needs" since it's not an external link. The other couple items I found are [here in Figma](https://www.figma.com/design/P1MgY68Y7Jju5PCfaJHQX4/Misc.-MFS-Stuff?node-id=803-25617&t=K2DvYD2ZDCzD8ZHe-1). The checkmark is more important to fix than the slight sizing and alignment issue of the attach icon. Otherwise, nothing else was jumping out at me.
Screenshot 2025-02-13 at 2 16 24 PM

kimmccaskill-oddball:
While Natasha mentioned that the sizing was more important, I still think we shouldn't introduce the alignment issues here  Here's the screenshot now that the icon is bigger
develop(before)
Screenshot 2025-02-17 at 10 14 55 AM

current branch(after)
Screenshot 2025-02-17 at 10 12 53 AM

matt-guest-wilcore:
I discussed with Natasha and updated the code to center align the icon and the text. I also adjusted the size of the paperclip icon to make it a bit smaller. I believe it looks a lot better now.

kimmccaskill-oddball (approved):
Looks good!

PR #10403, opened 2025-01-14
Reduce height of checkmark icon

kimmccaskill-oddball:
## Description of Change

The chevron height for the theme dimensions were set at 45px. This caused extra space when the contents to the left of the chevron were less than that. 
While there were multiple instances of the `chevronListItemHeight` with the chevron icon in the app, these before/after images were the ones I found that made a difference in the height. The reason others wouldn't change is because the content that precedes it is taller than the height of the chevon

## Screenshots/Video

| Before  | After |
| ------------- | ------------- |
| Screenshot 2025-01-14 at 10 40 37 AM  |  Screenshot 2025-01-14 at 10 47 53 AM |
| Screenshot 2025-01-14 at 10 42 27 AM | Screenshot 2025-01-14 at 10 47 14 AM  |
| Screenshot 2025-01-14 at 10 59 10 AM | Screenshot 2025-01-14 at 10 55 49 AM  |

## Testing

- [x] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


matt-guest-wilcore (approved):
looks good.

PR #10421, opened 2025-02-19
chore/192-UpgradeDesignSystemLibrary

oddballdave:
## Description of Change

Modified package.json to upgrade the design system components from 0.27.1 to 0.30.2 and the tokens from 0.20.0 to 0.23.0.

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## Reviewer Validations

Visually inspect all screens for correctness.

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/package.json
+++ b/VAMobile/package.json
@@ -40,8 +40,8 @@
   },
   "dependencies": {
     "@department-of-veterans-affairs/mobile-assets": "0.14.0",
-    "@department-of-veterans-affairs/mobile-component-library": "0.27.1",
-    "@department-of-veterans-affairs/mobile-tokens": "0.20.0",
+    "@department-of-veterans-affairs/mobile-component-library": "0.30.2",
+    "@department-of-veterans-affairs/mobile-tokens": "0.23.0",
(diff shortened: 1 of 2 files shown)


ATeal (approved):
Ty for getting this updated!!

VAMobile/package.json
```diff
@@ -81,7 +81,7 @@
     "react-native-safe-area-context": "^4.10.9",
     "react-native-screens": "^3.34.0",
     "react-native-shadow-2": "^7.1.0",
-    "react-native-svg": "^14.2.0",
+    "react-native-svg": "15.8.0",
```
kimmccaskill-oddball, line 84: Intentional caret removal?

ATeal: Good catch lets validate this

oddballdave: Yes the "no caret" is intentional. There are some compatibility issues with the DS Library if the versions don't match, so thought it was safer to lock it down like was done [here](https://github.com/department-of-veterans-affairs/va-mobile-library/blob/df7dc9e8d596cadffa2fdb7ecda2a763e2bc774a/packages/components/package.json#L44).

PR #10454, opened 2025-03-03
bug/9783-ForegroundNotificationsDoNotDisappear

oddballdave:
## Description of Change

Foreground notifications are notifications that are received while the app is running in the foreground. On Android, clicking on a foreground notification would correctly navigate the user to a deep link (e.g., Secure Message). However, after navigation the notification would remain in the notifications list. The notification list can be viewed by pulling down from the top of an Android device. This remaining notification acted strangely. Clicking on it did not launch the app or take the user to a deep link within the app.

To resolve this, [setAutoCancel(true)](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setAutoCancel(boolean)) was added to the [VAPushNotifications.kt](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/android/app/src/main/java/gov/va/mobileapp/notifications/VAPushNotifications.kt) module. This Android API cancels the notification when the user clicks on it.

## Testing

- [ ] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


ATeal (approved):
Nice find, I'm surprised thats not the default behavior

PR #10480, opened 2025-03-06
254 - Endpoint Configuration for vet_verification_status

matt-guest-wilcore:
## Description of Change

Added the new _vet_verification_status_ endpoint to the mobile app with unit tests. We can now successfully query the endpoint to check a veterans Title 38 status.

[Original Ticket](https://github.com/department-of-veterans-affairs/va-mobile-feature-support/issues/254)

## Testing

- [x] Tested on iOS 
- [ ] Tested on Android 

## Reviewer Validations

See ticket above for more detail. Verify the different responses and that data comes back as expected.

## PR Checklist

## For QA


VAMobile/src/api/types/VeteranVerificationStatusData.ts
```diff
@@ -0,0 +1,15 @@
+export type VeteranStatus = 'confirmed' | 'not confirmed'
+
+export type VeteranVerificationStatusDataAttributes = {
+  veteranStatus: VeteranStatus
+  notConfirmedReason?: string
+}
+
+export type VeteranVerificationStatusData = {
```
oddballdave, line 8: Would it be more consistent to name this VeteranVerificationStatusPayload? Seems like the app is a bit inconsistent here, but since this type has a data field maybe it's more of a payload?

VAMobile/src/api/veteranStatus/getVeteranStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { get } from 'store/api'
+
+import { veteranStatusKeys } from './queryKeys'
+
+/**
+ * Fetch veteran verification status
+ */
+const getVeteranStatus = async (): Promise<VeteranVerificationStatusData | undefined> => {
+  const response = await get<VeteranVerificationStatusData>('/v0/vet_verification_status')
+  if (response) {
```
oddballdave, line 14: Should this just be `return response` or `return response?.data` instead of spreading the response?

VAMobile/src/api/veteranStatus/getVeteranStatus.test.tsx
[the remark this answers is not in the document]
matt-guest-wilcore: Fair enough, I added tests because the ticket mentioned that unit tests should be updated and I figured I'd add something. But you’re right that testing the VeteranStatus screen directly or adding e2e would be way more valuable. In that case, I’m good with just dropping this test file for now. Once we flesh out Veteran Status a bit more, we can focus on better unit coverage and e2e tests.

VAMobile/src/api/types/VeteranVerificationStatusData.ts
matt-guest-wilcore: My thought process was to be consistent with the other data types. In the `ServiceHistoryData.ts` file we have the same data field so I used that as a reference, I've seen other files do this too. However, I'm ok with updating to VeteranVerificationStatusPayload.

oddballdave (approved):

PR #10481, opened 2025-03-07
Add travel pay entry point

juancstlm-a6:
## Description of Change

- Add the entry point to the travel pay SMOC flow using the same logic as vets website uses to determine if an appointment is eligible for travel pay. 
- Adds a line on the **Past** appointments list for any appointments that match the criteria.
- Adds a `AppointmentFileTravelPayAlert` component that renders only when the appointment matches the travel pay criteria

## Screenshots/Video
<details><summary>Before/after: </summary>

| Before | After | 
| --- | --- |
|Screenshot 2025-03-07 at 9 21 58 AMScreenshot 2025-03-07 at 9 22 08 AM|Screenshot 2025-03-07 at 9 09 37 AMScreenshot 2025-03-07 at 9 09 32 AM|
|Screenshot 2025-03-07 at 9 28 57 AMScreenshot 2025-03-07 at 9 28 51 AM|Screenshot 2025-03-07 at 9 27 03 AMScreenshot 2025-03-07 at 9 27 08 AM|
</details>

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


VAMobile/src/store/api/demo/mocks/appointments.json
```diff
@@ -187,7 +187,14 @@
             "reason": null,
             "isCovidVaccine": false,
             "serviceCategoryName": null,
-            "typeOfCare": "Mental Health"
+            "typeOfCare": "Mental Health",
```
liztownd, line 190: I know we really only care whether or not there's a travel claim appended to certain types of appointments, but for the record, the travelPayClaim will be appended to every appointment, regardless of type of care.

liztownd (approved):
Should probably have a mobile-specific person's input, but from a travel pay perspective it's looking good!

One thing to note is passing the include flag to the API call to retrieve the appts. I haven't spent much time looking at the Mobile module for appts, so can't confirm at this time, but we'll need to pass the `travel_pay_claim` in the `include` list to make sure the claims are appended when we get the appts.

VAMobile/src/api/types/AppointmentData.ts
```diff
@@ -105,6 +105,11 @@ export type AppointmentLocation = {
   code?: string
 }
 
+export type AppointmentTravelPayClaim = {
+  metadata: { status: number; success: boolean; message: string }
+  claim?: Record<string, unknown>
```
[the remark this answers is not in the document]
juancstlm-a6: Went ahead and added it.

juancstlm-a6:
> Should probably have a mobile-specific person's input, but from a travel pay perspective it's looking good!
> 
> One thing to note is passing the include flag to the API call to retrieve the appts. I haven't spent much time looking at the Mobile module for appts, so can't confirm at this time, but we'll need to pass the `travel_pay_claim` in the `include` list to make sure the claims are appended when we get the appts.

@liztownd 
Thanks for calling that out. I was going to tackle that portion in https://github.com/department-of-veterans-affairs/va.gov-team/issues/91570 I'll update it to include the details you shared here about passing the `travel_pay_claim` in the `include` param

liztownd:
@juancstlm-a6 Heads up, I mis-spoke, the flag is actually `travel_pay_claims` with an `s`. Double checked in the `vets-website` code where the flag is being passed in:

`includeParams.push('travel_pay_claims')`

VAMobile/src/api/types/AppointmentData.ts
```diff
@@ -105,6 +105,32 @@ export type AppointmentLocation = {
   code?: string
 }
 
+export type ClaimStatus =
+  | 'In Process'
```
liztownd, line 109: 2 things here:

1) these will be coming back from the API in `Sentence case` rather that `Title Case`

2) This honestly makes me nervous to make this a union type specifically calling out the existing statuses. If VTP adds a new status (which they have - they only recently added `On hold`) it will break the Mobile app when a new, undefined status comes through. Might be safer to leave this just a `string` type. (Not my favorite thing ever, but I'd rather be safe than sorry.)

VAMobile/src/api/types/AppointmentData.ts
```diff
@@ -105,6 +105,32 @@ export type AppointmentLocation = {
...
+  | 'Appealed'
+  | 'In Manual Review'
+  | 'Closed'
+  | 'On Hold'
+  | 'Denied'
+
+export type TravelPayClaim = {
+  id: string
+  claimNumber: string
+  claimStatus: 'In Process'
```
liztownd, line 122: This should be `ClaimStatus`.

But honestly, I'd rather keep it just a `string` type to future-proof against new statuses being added by VTP (see comment above).

VAMobile/src/api/types/AppointmentData.ts
juancstlm-a6: Given that I think it makes sense to just leave it as a string type for now.

VAMobile/src/api/types/AppointmentData.ts
juancstlm-a6: A good catch. Looks like I missed staging this change and it didn't make it through. Updated to just a string type for now

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/InPersonVAAppointment.tsx
```diff
@@ -44,22 +38,12 @@ function InPersonVAAppointment({
...
         <AppointmentDateAndTime attributes={attributes} subType={subType} />
-        {subType === AppointmentDetailsSubTypeConstants.Past && (
-          <Box>
-            <Button
-              onPress={() => navigateTo('SubmitTravelPayClaimScreen', { appointmentDateTime: attributes.startDateUtc })}
-              label="Submit travel claim"
-              a11yLabel="Submit travel claim"
-            />
-          </Box>
-        )}
```
dumathane, line 62: Why all this removal?

VAMobile/src/store/api/demo/mocks/appointments.json
```diff
@@ -40,7 +40,14 @@
...
             "reason": "Test",
             "isCovidVaccine": false,
-            "typeOfCare": "Podiatry"
+            "typeOfCare": "Podiatry",
+            "travelPayClaim": {
+              "metadata": {
+                "status": 200,
+                "message": "No claims found.",
+                "success": true
+              }
```
dumathane, line 49: Aren't all of these missing the actual claim data?

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/InPersonVAAppointment.tsx
dumathane: I think I get it now: Removing some older code and replacing this by not specifying the appt types specifically and instead doing it in the PastAppointmentDetails and relying a on a Nil return if they don't fit the criteria? Correct me if I'm wrong?

VAMobile/src/store/api/demo/mocks/appointments.json
dumathane: Is this just a step in the right direction and the actual full claim detail is to come next?

dumathane (changes requested):
I think this all looks pretty great, left a few notes I'd just like a little clarification on. Marking request changes until I hear back but then will approve.

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/InPersonVAAppointment.tsx
juancstlm-a6: Correct. This was older scaffolding code that Liz added to get into the feature w/o having any of the actual logic in place to build out the screens

VAMobile/src/store/api/demo/mocks/appointments.json
juancstlm-a6: Correct. We are not really using the details at the moment and just like web only care if there is a claim or not.

dumathane: You don't currently have any sort of 'claim already submitted' or 'claim paid' indicator yet though right? That is probably also to come in the future?

dumathane: Any thoughts on indicating if someone is past the claim window to them as well like 'hey you could have but now it's too late' type situation?

juancstlm-a6: Yes, this will be done as part of another ticket. The changes in scope here are just to remove the existing feature entry point and replace it with one that includes the logic such that it matches what exists on web.

juancstlm-a6: > You don't currently have any sort of 'claim already submitted' or 'claim paid' indicator yet though right? That is probably also to come in the future?

This is all part of future tickets

dumathane (approved):

PR #10509, opened 2025-03-11
282 - Updating .gitignore

matt-guest-wilcore:
## Description of Change

Updating the `.gitignore` file to exclude the `.xcode.env.local` file as discussed in our MFS weekly engineering sync. In my own personal workflow the `.xcode.env.local` is the only one I have to manually keep track of to get the iOS app running.

If there are any other files we should ignore please make sure to mention them.

[Original Ticket](https://github.com/department-of-veterans-affairs/va-mobile-feature-support/issues/282)

## Testing

Verify that your .xcode.env file is no longer tracked

- [x] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA


VAMobile/ios/.xcode.env
```diff
@@ -1,10 +0,0 @@
-# This `.xcode.env` file is versioned and is used to source the environment
-# used when running script phases inside Xcode.
-# To customize your local environment, you can create an `.xcode.env.local`
```
timwright12, line 3: @matt-guest-wilcore have you tried using the recommended .xcode.env.local file?

matt-guest-wilcore: I haven't, I've been using `.xcode.env` this entire time 😅 . In any case, based on discussions my team had some time ago we wanted to update the .gitignore so we didn't have to worry about local setup files being tracked. Seems like I should add the `.xcode.env.local` file to the gitignore instead.

.gitignore
```diff
@@ -122,6 +122,7 @@ coverage
 **/ios/vendor
 **/ios/Pods
 **/ios/main.jsbundle.map
+/VAMobile/ios/.xcode.env.local
```
timwright12, line 125: Can you run a test to see what happens with this file deleted? If new team members come onboard and it turns out this file is required somehow we'll need the docs updated as well

kimmccaskill-oddball: Would a good work around be to add the imports into `.xcode.env`? Sorry if thats a silly question. I'm not super familiar with these files

matt-guest-wilcore: @timwright12 My understanding is that the file is important and is needed to get the app running, so a new team member would need to add `.xcode.env.local` to their project with all the imports and stuff to get things running. I just tested it myself and after adding the file and doing my usual setup things are working fine, I can see that it's not being tracked either.

timwright12: If it's needed, then I'd like to keep it out of the gitignore. I'm not sure why that file is giving you problems, but maybe we can look at the workflow because I think this is the first time in 4 years it's been a problem

kimmccaskill-oddball: Here's the exact error I'm getting that causes the build to fail. Running the command in this suggestion that the warning output below is what sparked edits to this file initially
Screenshot 2025-03-17 at 10 20 26 AM

PR #10575, opened 2025-03-20
Enable all users to view veteran status card

matt-guest-wilcore:
## Description of Change

Updated the `Nametag` component to remove its reliance on honorable-discharge checks before showing the `VeteranStatusScreen`. We now always allow users to tap through to the Veteran Status screen, where Title 38 eligibility is handled. Keep in mind, the user still has to have a service history in order to see the `Nametag` component.

[Original ticket](https://github.com/department-of-veterans-affairs/va-mobile-feature-support/issues/260)

## Screenshots/Video

Tested on user 32 (Tamara Ellis).

vets.gov.user+32@gmail.com
134SsNrLgPv5

https://github.com/user-attachments/assets/e9104eb9-529f-4cb7-9253-0989ec87e2c7

## Testing

Account with service history & title 38 confirmed:
[vets.gov](http://vets.gov/).[user+16@gmail.com](mailto:user+16@gmail.com)
296SsNrLgPv5

[Other test accounts for Veteran Status (YMMV)](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/Administrative/vagov-users/staging-test-accounts-Veteran-Status.md)

**Note:** Because most staging accounts lack any service history, you won’t see the Nametag (and thus the VeteranStatusScreen) by default—even if a user is Title 38 eligible. If you temporarily mock service history data in the Nametag (ensuring branch isn’t undefined), you’ll be able to trigger the VeteranStatusScreen and view the relevant messages for testing.

- [X] Tested on iOS 
- [ ] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/components/Nametag/Nametag.tsx
+++ b/VAMobile/src/components/Nametag/Nametag.tsx
@@ -33,7 +33,7 @@ export const Nametag = () => {
     }
   }, [personalInfo])
 
-  const accLabel = branch !== '' ? `${branch} ${t('veteranStatus.proofOf')}` : branch
+  const accLabel = branch !== '' ? `${branch} ${t('veteranStatus.proofOf')}` : undefined
 
   const pressableProps: PressableProps = {
     onPress: () => navigateTo('VeteranStatus'),
--- a/VAMobile/src/components/Nametag/Nametag.tsx
+++ b/VAMobile/src/components/Nametag/Nametag.tsx
@@ -8,6 +8,8 @@ import { colors } from '@department-of-veterans-affairs/mobile-tokens'
 import { useAuthorizedServices } from 'api/authorizedServices/getAuthorizedServices'
 import { useServiceHistory } from 'api/militaryService'
 import { usePersonalInformation } from 'api/personalInformation/getPersonalInformation'
+import { BranchOfService } from 'api/types'
+import { useVeteranStatus } from 'api/veteranStatus'
 import { BackgroundVariant, Box, MilitaryBranchEmblem, TextView } from 'components'
 import { UserAnalytics } from 'constants/analytics'
 import { NAMESPACE } from 'constants/namespaces'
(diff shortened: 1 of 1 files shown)


dumathane:
> Keep in mind, the user still has to have a service history in order to see the Nametag component.

Feels like this is in direct contradiction to the comments I added on your other PR with the error messaging. If the nametag will only show up when we have a service history then why are we even checking for it in the vet status card page. Feels like we should always display vet status button and then have the error messages available on the page itself regarding their history missing?

VAMobile/src/components/Nametag/Nametag.tsx
```diff
@@ -33,28 +33,19 @@ export const Nametag = () => {
...
-    (service) => service.honorableServiceIndicator === 'Y',
-  )
-
-  let accLabel
-  if (!accessToMilitaryInfo) {
-    accLabel = undefined
-  } else {
-    accLabel = showVeteranStatus ? `${branch} ${t('veteranStatus.proofOf')}` : branch
-  }
+  const accLabel = branch !== '' ? `${branch} ${t('veteranStatus.proofOf')}` : branch
```
dumathane, line 36: This is a little misleading...can you make the `: branch` into a `: undefined`, as I think that is more accurate as to what's happening here.

matt-guest-wilcore:
@dumathane I discussed with my team and IIR yesterday and it was decided that we want to keep the original behavior for non-Veterans. The NameTag component relies on service history and branch info for the emblem and text, so we hide it if that data doesn’t exist. That doesn’t contradict the new error flow in the VeteranStatusScreen—it just preserves the existing UX for users with no military service data. If that changes in the future, we can revisit the logic.

As for the other PR you mentioned, we check service history because there is an edge case where a user is "confirmed" by the vets verification status but also has no service history. The Figma shows this [here](https://www.figma.com/design/AobGKkBvIiDuYXR2QqkytC/IIR-Veteran-Status?node-id=641-84833&t=tfiJCged5yyhwwOe-0). I hope this all makes sense.

VAMobile/src/components/Nametag/Nametag.tsx
matt-guest-wilcore: Sure, i can make that update

dumathane:
> That doesn’t contradict the new error flow in the VeteranStatusScreen—it just preserves the existing UX for users with no military service data. If that changes in the future, we can revisit the logic.
> The Figma shows this [here](https://www.figma.com/design/AobGKkBvIiDuYXR2QqkytC/IIR-Veteran-Status?node-id=641-84833&t=tfiJCged5yyhwwOe-0). I hope this all makes sense.

I think you're still missing my point here. Isn't the veteran status screen 'no service history' error largely impossible to get to if we don't show the nametag component when the service history length is nada? I'm fine with a redundant check here, but that's largely what I was getting at.

matt-guest-wilcore:
Ah I think I see what you mean. Yes, technically we would never see the "no service history" error in the VeteranStatusScreen because if the user had no service history they won't see the Nametag in the first place.

dumathane (approved):

matt-guest-wilcore:
@dumathane After some additional discussion with my team, we decided to add one more check. We still hide the Nametag for non-veterans, but now if someone is ‘confirmed’ and has no service history, they’ll still see the Nametag. That way, they can tap through to VeteranStatusScreen and see an error with actionable steps. We found cases where a user can be confirmed in the new endpoint but not show any service history data, so that needed to be handled.

I tested with multiple users (confirmed/not, with/without service history), and everything behaves as intended. The UI can look a bit bare if the user has no branch data, but we can handle that in the upcoming VeteranStatus redesign work. Here's an example of how it looks for a user who is "confirmed" with no service history.

dumathane (approved):
That last edit makes total sense to me, I'd rather let them hit some sort of message than just hide the functionality silently.

PR #10596, opened 2025-03-24
103595 add e2e test for submit travel pay flow  happy path

juancstlm-a6:
## Description of Change
- Adds E2E test for the Travel Pay SMOC flow
- Updates appointments expanded test to check for travel pay alert

In order to accommodate the travel pay smoc flow for a user that has an address, as the demo user doesn't have one, I opted to add a function that adds an address before the travel pay flow e2e tests so that when the flow is triggered it can flow through the happy path w/o issues.

Due to having borrowed constants from the `ContactInformation.e2e.ts` file this was causing issues with the test as importing the constants causes the `beforeAll` function in the `ContactInformation.e2e.ts` to be executed along with all the test in that file. I opted to move the constants to the `utils` file so that the set up and tests in `ContactInformation.e2e.ts` don't interfere with the new `TravelPay` tests

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


VAMobile/e2e/tests/AppointmentsExpanded.e2e.ts
```diff
@@ -755,7 +773,7 @@ export async function apppointmentVerification(pastAppointment = false) {
     }
     await checkUpcomingApptDetails(
       'Claim',
-      'Upcoming',
```
juancstlm-a6, line 758: I think this should be changed to 'Confirmed' as the test here specifies claim exam confirmed but the argument is `Upcoming` 

Test pass with this change

VAMobile/e2e/tests/TravelPay.e2e.ts
```diff
@@ -0,0 +1,235 @@
...
+  await expect(element(by.id(CommonE2eIdConstants.CALL_VA_PHONE_NUMBER_ID))).toExist()
+  await expect(element(by.id(CommonE2eIdConstants.CALL_VA_TTY_PHONE_NUMBER_ID))).toExist()
+}
+
+const checkAddressScreen = async (existingAddress: boolean) => {
+  await expect(element(by.id(TravelPayE2eIdConstants.ADDRESS_QUESTION_ID))).toExist()
+  if (existingAddress) {
+    await expect(element(by.id(TravelPayE2eIdConstants.ADDRESS_QUALIFIER_ID))).toExist()
+    await expect(element(by.id(TravelPayE2eIdConstants.REFER_TO_PORTAL_ID))).toExist()
+    await expect(element(by.id('Home address 3101 N Fort Valley Rd Flagstaff, AZ, 86001'))).toExist()
```
liztownd, line 178: `by.id` ? I'm not sure I understand this yet. I feel like it should be `by.text` maybe?

VAMobile/e2e/tests/ContactInformation.e2e.ts
```diff
@@ -3,75 +3,44 @@ import { setTimeout } from 'timers/promises'
...
-  VERIFY_YOUR_ADDRESS_ID: 'verifyYourAddressTestID',
-  EMAIL_ADDRESS_EDIT_ID: 'emailAddressEditTestID',
-  CONTACT_INFO_CLOSE_ID: 'ContactInfoCloseTestID',
-  MILITARY_POST_OFFICE_PICKER_CONFIRM_ID: 'militaryPostOfficeConfirmID',
-  HOW_WE_USE_CONTACT_INFO_LINK_ID: 'howWeUseContactInfoLinkTestID',
-}
-
 export async function updateAddress() {
-  await waitFor(element(by.id(ContactInfoE2eIdConstants.COUNTRY_PICKER_ID)))
+  await waitFor(element(by.id(CommonE2eIdConstants.COUNTRY_PICKER_ID)))
```
liztownd, line 7: I don't know enough about the mobile detox tests to understand the difference between the `ContactInfoE2eIdConstants` and the `CommonE2eIdConstants` here. Will rely on a Mobile team member for this bit....

Ok, I think I understand now - they were moved to the utils file so they can be shared, I think....

VAMobile/e2e/tests/ContactInformation.e2e.ts
```diff
@@ -3,75 +3,44 @@ import { setTimeout } from 'timers/promises'
 
 import { CommonE2eIdConstants, loginToDemoMode, openContactInfo, openProfile, toggleRemoteConfigFlag } from './utils'
 
-export const ContactInfoE2eIdConstants = {
```
liztownd, line 6: Again, not sure I have the necessary knowledge to understand the implications of removing this here....

Edit: I think I understand - they were moved to the utils file so they can be shared, yeah?

VAMobile/e2e/tests/utils.ts
```diff
@@ -146,6 +146,35 @@ export const CommonE2eIdConstants = {
   AF_ERROR_MSG_PHONE_ID: 'AFErrorPhoneNumberTestID',
   AF_TYPE_INPUT_ID: 'AFTypeTestID',
   AF_USE_CASE_TWO_ID: 'AFUseCase2TestID',
+  // Contact information
+  MAILING_ADDRESS_ID: 'Mailing address 3101 N Fort Valley Rd Flagstaff, AZ, 86001',
```
liztownd, line 150: Ah, I see, this was moved from the contact information test file to here.... so it can be shared and overwritten for the travel pay test. Gotchya. 👍

VAMobile/e2e/tests/utils.ts
```diff
@@ -185,7 +214,7 @@ export async function loginToDemoMode(skipOnboarding = true, pushNotifications?:
   }
 
   await element(by.id(CommonE2eIdConstants.DEMO_MODE_INPUT_ID)).tapReturnKey()
-  await element(by.id(CommonE2eIdConstants.DEMO_BTN_ID)).multiTap(2)
+  await element(by.id(CommonE2eIdConstants.DEMO_BTN_ID)).multiTap(1)
```
liztownd, line 217: Confused by this change. 🤔

VAMobile/src/screens/HealthScreen/TravelPay/SubmitTravelPayFlowSteps/ReviewClaimScreen.tsx
```diff
@@ -28,27 +28,37 @@ function ReviewClaimScreen({}: ReviewClaimScreenProps) {
...
-        <TextView variant="MobileBody" mt={theme.dimensions.standardMarginBetween}>
+        <TextView testID="reviewTextID" variant="MobileBody" mt={theme.dimensions.standardMarginBetween}>
           {t('travelPay.reviewText')}
         </TextView>
         <Box mt={theme.dimensions.standardMarginBetween}>
           <TextArea>
-            <TextView variant="MobileBodyBold">{t("What you're claiming")}</TextView>
-            <TextView variant="MobileBody">{t('Mileage only')}</TextView>
+            <TextView testID="whatID" variant="MobileBodyBold">
+              {t("What you're claiming")}
```
liztownd, line 40: Weren't these text strings moved to the translations file? (Or is that in a PR that hasn't been merged yet?)

VAMobile/e2e/tests/utils.ts
juancstlm-a6: This might not be needed and should have not been stage. My bad here. Without this change the detox test would not run for me locally. It would trigger 2 taps on the demo button. The first tap would enter demo mode and clear the input field and the second would try to enter demo mode but since the input was cleared the password would not match and would error out the entire tests. Not sure how it's working fine for the git hub actions flow but it wasn't working for me w/o this change.

TLDR: Needs to be reverted, only used to run e2e test locally for me.

VAMobile/src/screens/HealthScreen/TravelPay/SubmitTravelPayFlowSteps/ReviewClaimScreen.tsx
juancstlm-a6: I'll have to update this once the other PR's are merged or update this PR branch with those changes

VAMobile/e2e/tests/TravelPay.e2e.ts
juancstlm-a6: Nice, catch. Not sure how they passed before but this should be `by.text` like you mentioned

juancstlm-a6: Actually this should be an `by.id` since the component that render this joins all the lines together and sets it as the test id

Issue #10595, opened 2025-03-24
BUG - [SEV 3] - [Android - Samsung only] - [Screenreader pronounces "Vaw" and "Via" rather than "VA" on new messaages / Rx error screens]


digitalken:
## What happened?

Depending on reading mode (whole page, partial page), Pixel phones read "Vaw" or "Via" rather than "VA" on the new Messages and Prescriptions access error pages.  See screen shots.
## Specs:

- Device: Samsung
- OS: Android
- User Account:
- Accessibility State: Screen reader

## Steps to Reproduce

User screen reader to read the error pages.

## Desired behavior

Should pronounce occurences of "VA" as "VA" rather than "Vaw" or "Via".
## Bug Severity - BE SURE TO ADD THE SEVERITY LABEL
 
See [Bug Tracking](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/QA#issue-severity) for details on severity levels

- Impact: Low 
- Frequency:  Low 

- 3 - Low

## Linked to Story

#10501

## Screen shot(s) and additional information

![Image](https://github.com/user-attachments/assets/2a178bfb-9dad-4a3c-8f17-23b96e3ea417)

<details>
  <summary>Full JSON response for services related to issue (expand/collapse)</summary>
  

</details>

## Ticket Checklist
- [ ] Steps to reproduce are defined
- [ ] Desired behavior is added
- [ ] Labels added (front-end, back-end, global, Health, relevant feature, accessibility, etc)
- [ ] Estimate of 1 added (for front-end tickets)
- [ ] Added either to the relevant feature epic (if found during new feature implementation), or the relevant team's bug epic ([Global](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8302), [Health & Benefits](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8303), [API](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8304), [QART](https://app.zenhub.com/workspaces/va-mobile-60f1a34998bc75000f2a489f/issues/gh/department-of-veterans-affairs/va-mobile-app/8305))

digitalken:
[...]

PR #10637, opened 2025-03-28
Android/Samsung - VA pronunciation on Rx error screen

jmarchi:
## Description of Change
Ticket: https://github.com/department-of-veterans-affairs/va-mobile-app/issues/10595

Added util call to fix pronunciation on RX error screen copy that contained "VA" but was not using the util

## Testing
Checked on a Pixel, but I don't have a Samsung available so will work with @digitalken on the QA pass to confirm the fix worked

- [X] Tested on iOS 
- [X] Tested on Android 

## PR Checklist

## For QA


keli13 (dismissed):

dftony:
Weird that a Pixel phone read VA correctly without these helpers

dftony (approved):

timwright12 (dismissed):

dumathane (approved):

digitalken (approved):
This works well with other screen readers and there has been -some- improvement to the Samsung screen reader (it now reads "via" rather than "vaw" in a few places) in my opinion. Samsung reads this screen similarly to the way it reads other screens in the app, currently.

PR #10639, opened 2025-03-29
Add alternate paths

juancstlm-a6:
## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA


VAMobile/e2e/tests/TravelPay.e2e.ts
```diff
@@ -93,27 +90,17 @@ const fillHomeAddressFields = async () => {
   await element(by.id(CommonE2eIdConstants.CONTACT_INFO_SAVE_ID)).tap()
   await waitFor(element(by.id(CommonE2eIdConstants.CONTACT_INFO_SUGGESTED_ADDRESS_ID)))
     .toBeVisible()
-    .withTimeout(2000)
+    .withTimeout(4_000)
```
liztownd, line 93: I'm confused by these underscores. 🤔 (Especially when I noticed in some later tests that the timeout is set without the underscore)

liztownd (approved):
Had some syntax questions, but looks good!

VAMobile/e2e/tests/TravelPay.e2e.ts
juancstlm-a6: It's just a different way to say the time. It's supposed to be more readable but I used it as a way to re-trigger a test run w/o changing much. Reverted them all back to match the existing non underscore times

PR #10659, opened 2025-04-03
Travel Pay - Add remote config

juancstlm-a6:
## Testing

- [x] Tested on iOS 
- [x] Tested on Android 

## PR Checklist

## For QA

### DIFF
--- a/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert.test.tsx
+++ b/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert.test.tsx
@@ -198,7 +198,7 @@ context('AppointmentFileTravelPayAlert', () => {
       expect(screen.queryByTestId('appointmentFileTravelPayAlert')).toBeNull()
     })
 
-    it('should NOT render if no travel pay claim', async () => {
+    it('should NOT render if unable to retrieve travel pay claim data', async () => {
       const attributes = createTestAppointmentAttributes({
         status: AppointmentStatusConstants.BOOKED,
         appointmentType: AppointmentTypeConstants.VA,
--- a/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert.test.tsx
+++ b/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert.test.tsx
@@ -12,7 +12,10 @@ import {
   AppointmentTypeConstants,
 } from 'api/types'
 import AppointmentFileTravelPayAlert from 'screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert'
-import { context, fireEvent, render, screen } from 'testUtils'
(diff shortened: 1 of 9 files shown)


VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert.test.tsx
```diff
@@ -169,126 +172,136 @@ const tests = [
...
 
-  it('should render if zero days left to file', async () => {
-    const attributes = createTestAppointmentAttributes({
-      status: AppointmentStatusConstants.BOOKED,
-      appointmentType: AppointmentTypeConstants.VA,
-      startDateUtc: DateTime.now().minus({ days: 30 }).toUTC().toISO(),
-      isPending: false,
-      phoneOnly: false,
-      travelPayClaim: travelPayClaimData,
+    it('should NOT render if no travel pay claim', async () => {
```
liztownd, line 201: Just a note, this was confusing to me. 

Might consider renaming the test to indicate that we're unable to determine if a claim has already been filed vs. just "no travel pay claim" (the fact that there's no travel claim initially made me think that we should definitely render the link to file, but it's actually not rendering because we were unable to retrieve travel claim information for the appointment at all).

liztownd (approved):
I think this looks good - with one caveat:

I honestly don't have a great understanding of integrating with the Firebase Remote Config, so would love to have a second pair of (Mobile-team-specific) eyes on this just to confirm.

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentFileTravelPayAlert.tsx
```diff
@@ -18,6 +19,10 @@ function AppointmentFileTravelPayAlert({ attributes }: AppointmentFileTravelPayA
   const theme = useTheme()
   const navigateTo = useRouteNavigation()
 
+  if (!featureEnabled('travelPaySMOC')) {
```
dumathane, line 22: Can we move this to PastAppointmentDetails.tsx layer. Something like this:

{featureEnabled('travelPaySMOC') && (<AppointmentFileTravelPayAlert attributes={attributes} />)}

That's the pattern we've used elsewhere for stuff like this. It'll help when you start enabled analytics as well I think.

VAMobile/src/utils/appointments.tsx
```diff
@@ -372,6 +373,10 @@ export const getDaysLeftToFileTravelPay = (startDateUtc: string) => {
 }
 
 const getTravelPay = (attributes: AppointmentAttributes, t: TFunction, mb: number) => {
+  if (!featureEnabled('travelPaySMOC')) {
+    return undefined
+  }
+
```
dumathane, line 379: Same here. Make the call up a level to avoid the function call altogether and not to return blank upon the call.

dumathane (changes requested):
Move featureEnabled calls

keli13:
@juancstlm-a6 in terms of front end changes for remote config - this looks good to me and if @dumathane is good with the recent changes.

In terms of integration with firebase, will have manually add the new config to firebase when we are ready to go live and coordinate that.

You can test your new remote config locally though by logging into demo mode and toggle the option in the developer screen(Profile -> Settings -> Developer Screen -> Remote Config -> hit apply after toggle change).

keli13 (approved):

PR #10833, opened 2025-05-07
107847 claim not filed

juancstlm-a6:
## Description of Change
- Adds content to show when a claim was not filed within 30 days

## Link to Issue
https://github.com/department-of-veterans-affairs/va.gov-team/issues/107847

## Screenshots/Video

<details><summary>Toggle</summary>
Screenshot 2025-05-06 at 11 35 40 PM
Screenshot_1746599835

</details>

## Testing Requirements

Acceptance Criteria

- [ ] Matches visual design in Figma (layout, typography, spacing, colors)
- [ ] Accessible according to A11y standards
- [ ] Unit tests
- [ ] E2E tests

Test Users

- [ ] Demo mode

## Checklist for PR Submitter

- [ ] Code is attached to a feature flag, or reason is given for no feature flag
- [ ] AC and Tests are included to cover this change (when possible)

## Checklist for QA

- [ ] Tested on iOS
- [ ] Tested on Android
- [ ] [Run a build for this branch](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/workflows/on_demand_build.yml)

### DIFF
--- a/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentTravelClaimDetails.tsx
+++ b/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentTravelClaimDetails.tsx
@@ -8,7 +8,12 @@ import { NAMESPACE } from 'constants/namespaces'
 import { VATheme } from 'styles/theme'
 import { a11yLabelVA } from 'utils/a11yLabel'
 import { logAnalyticsEvent } from 'utils/analytics'
-import { AppointmentDetailsSubType, AppointmentDetailsSubTypeConstants } from 'utils/appointments'
+import {
+  AppointmentDetailsSubType,
+  AppointmentDetailsSubTypeConstants,
+  appointmentMeetsTravelPayCriteria,
+  getDaysLeftToFileTravelPay,
+} from 'utils/appointments'
 import getEnv from 'utils/env'
 import { displayedTextPhoneNumber } from 'utils/formattingUtils'
 import { useRouteNavigation, useTheme } from 'utils/hooks'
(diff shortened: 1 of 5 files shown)


VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentTravelClaimDetails.tsx
```diff
@@ -89,6 +94,16 @@ function TravelClaimFiledDetails({ attributes, subType }: TravelClaimFiledDetail
       )
     }
 
+    const daysLeftToFileTravelPay = getDaysLeftToFileTravelPay(attributes.startDateUtc)
+
+    if (!claim && appointmentMeetsTravelPayCriteria(attributes) && daysLeftToFileTravelPay < 0) {
+      return (
```
liztownd, line 100: I'm wondering if this wouldn't be better in the FileTravelPayAlert component - since there you'd already be doing the eligibility checks and checking the days since the appointment, so it would be a simple "if eligible && < 30 days, show the alert with the link for smoc, if > 30 days show the you didn't file component". Thoughts?

That way this component has not filing related logic to it, only whether or not there's an existing claim, yeah?

liztownd (changes requested):
I recommend moving the "you didn't file within the time frame" component to within the alert component so they can share the eligibility and days since appointment logic, and just render either the "you're out of time" or "here's your link" depending on the time since the appointment.

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentTravelClaimDetails.tsx
juancstlm-a6: I like that idea. Only concern is that on the designs the "Did not file" replaces the "Details" component. If we are ok with the "Did not file" being at the top of the appointment details since that is where the Alert lives I would much rather consolidate the logic there.

liztownd: Ah, yeah, I see... 🤔  Yeah, I don't think we want it at the top of the appointment details. I wonder how else we might simplify all that logic then. There are just so many considerations to keep track of... 🫠   Let me think on it some more.

liztownd: Given the concern about placement of this component, I think this approach works out.

liztownd (approved):
Been thinking on the logic and component placement for the claim not filed component, and I think this is probably the best approach (since we don't want it at the top where the alert would be). I withdraw my request for changes. 😁

PR #10863, opened 2025-05-13
Claim Submission follow up

juancstlm-a6:
## Description of Change
- Modifies the cache data using only the `id` of the appointment 
- Updates the Navigation params of the `PastAppointmentDetails` component to include new claim data
- Updates e2e and unit test
- Adds gating of travel pay details component behind the feature toggle 

## Link to Issue
[](https://github.com/department-of-veterans-affairs/va.gov-team/issues/109545)

## Screenshots/Video
N/A

## Testing Requirements
- Verified using test both unit and e2e as well as going through the flow using demo mode on simulators

Acceptance Criteria

- [ ] After submitting a travel claim and closing the travel pay flow, the travel pay file alert is no longer visible.
- [ ] After submitting a travel claim and closing the travel pay flow, the travel pay details component is displayed.
- [ ] On submission mutation, the appointment is identified using the `id` attribute.
- [ ] When updating the appointments cache, all relevant time frames contain the claim data for the same appointment.
- [ ] Unit and end-to-end (e2e) tests are updated accordingly.
Test Users

- [ ] Demo Mode

## Checklist for PR Submitter

- [x] Code is attached to a feature flag, or reason is given for no feature flag
- [x] AC and Tests are included to cover this change (when possible)

## Checklist for QA

- [ ] Tested on iOS
- [ ] Tested on Android
- [ ] [Run a build for this branch](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/workflows/on_demand_build.yml)

### DIFF
--- a/VAMobile/src/api/travelPay/submitClaim.tsx
+++ b/VAMobile/src/api/travelPay/submitClaim.tsx
@@ -35,12 +35,7 @@ export const useSubmitTravelClaim = () => {
       const newAppointmentsList = oldAppointments.data.map((appointment) => {
         const newAppointment = { ...appointment }
 
-        //TODO: Find a better way to identify the appointment and append the claim data
-        if (
-          newAppointment.attributes.startDateUtc === variables.appointmentDateTime &&
-          newAppointment.attributes.location.id === variables.facilityStationNumber &&
-          !newAppointment.attributes.travelPayClaim?.claim?.id
-        ) {
+        if (newAppointment.id === variables.appointmentID) {
           newAppointment.attributes.travelPayClaim = {
             metadata: {
               status: 200,
--- a/VAMobile/src/api/travelPay/submitClaim.tsx
+++ b/VAMobile/src/api/travelPay/submitClaim.tsx
@@ -1,12 +1,12 @@
-import { useMutation, useQueryClient } from '@tanstack/react-query'
+import { QueryKey, useMutation, useQueryClient } from '@tanstack/react-query'
 
 import { appointmentsKeys } from 'api/appointments'
 import { AppointmentsGetData, SubmitSMOCTravelPayClaimParameters, SubmitTravelPayClaimResponseData } from 'api/types'
-import { TimeFrameTypeConstants } from 'constants/appointments'
+import { TimeFrameType, TimeFrameTypeConstants } from 'constants/appointments'
 import { Params as APIParams, post } from 'store/api/api'
 import { logNonFatalErrorToFirebase } from 'utils/analytics'
 import { isErrorObject } from 'utils/common'
-import { stripTZOffset } from 'utils/travelPay'
+import { appendClaimDataToAppointment, stripTZOffset } from 'utils/travelPay'
 
 const submitClaim = async (smocTravelPayClaimData: SubmitSMOCTravelPayClaimParameters) => {
   const endpoint = '/v0/travel-pay/claims'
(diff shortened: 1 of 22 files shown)


VAMobile/src/screens/HealthScreen/TravelPay/SubmitTravelPayFlowSteps/ReviewClaimScreen.tsx
```diff
@@ -79,14 +85,24 @@ function ReviewClaimScreen({ route }: ReviewClaimScreenProps) {
...
         appointmentType: 'Other',
         isComplete: false,
       },
       {
         onSuccess: (_data) => {
-          //TODOD: Modify the nav params to include the claim data
+          const appointments = queryClient.getQueryData([
+            appointmentsKeys.appointments,
+            TimeFrameTypeConstants.PAST_THREE_MONTHS,
+          ]) as AppointmentsGetData
```
juancstlm-a6, line 99: This data should always exist. Thoughts on if we should make this code a bit more defensive in the case it does not.

VAMobile/src/api/travelPay/submitClaim.tsx
juancstlm-a6, line 1: We should not need to update other appointment key time frames if the limit on submission is 30 days from appointment start time. Ass the appointment data is paginated in intervals and at current time a SMOC claim should only be submitted for those that fall under the `TimeFrameTypeConstants.PAST_THREE_MONTHS` key

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/ClaimExamAppointment.tsx
```diff
@@ -66,7 +66,7 @@ function ClaimExamAppointment({
         <AppointmentReasonAndComment attributes={attributes} subType={subType} type={type} />
         <AppointmentMedicationWording subType={subType} type={type} />
         <AppointmentPersonalContactInfo attributes={attributes} subType={subType} />
-        <AppointmentTravelClaimDetails attributes={attributes} subType={subType} />
+        {featureEnabled('travelPaySMOC') && <AppointmentTravelClaimDetails attributes={attributes} subType={subType} />}
```
liztownd, line 69: Rather than having this flag in each of the appointment types, would it be easier to manage if it was in the component itself to return `null` if the flag is not enabled? That way we only need to manage it in one location?

VAMobile/src/screens/HealthScreen/TravelPay/SubmitTravelPayFlowSteps/ReviewClaimScreen.tsx
liztownd: 🤔 Yeah, _probably_ it should exist. Maybe add `?` null chain coalescing to the snippet below to catch it? Or add an `if` statement, `if (appointments) {do whatever}` just in case? 

On the other hand, if the query data for whatever reason doesn't exist, once going back to the appointment details it _should_ make the API call again. So if it's not there we just skip this step, because it would make the appts call again and we'd then have updated data.... Right?

VAMobile/src/api/travelPay/submitClaim.tsx
liztownd: Fair. I do wonder about the `PAST_ALL_CURRENT_YEAR` - that's the only other one I can see having appts within the last 30 day time frame. But yeah, you're right, the rest of the constants won't have the time frame necessary to be filing a claim.

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/ClaimExamAppointment.tsx
juancstlm-a6: It would and I would rather do that. I believe when I did this for the TravelPayAlert component I was told  by the mobile team to follow the pattern above instead. Happy to change it back

liztownd: 🤔 Well, that complicates things. 🤣  I'd follow the advice of the mobile team, or maybe reach out to them and verify that this is the best approach.

VAMobile/src/screens/HealthScreen/TravelPay/SubmitTravelPayFlowSteps/ReviewClaimScreen.tsx
juancstlm-a6: I refactored to pass in the entire appointment data rather than query the data so that we always have the appointment data and just append the claim details

VAMobile/src/api/travelPay/submitClaim.tsx
juancstlm-a6: You're right I updated the way we mutate the data such that it looks through all appointment queries

VAMobile/src/api/travelPay/submitClaim.tsx
```diff
@@ -27,43 +27,41 @@ export const useSubmitTravelClaim = () => {
...
-              status: 200,
-              message: 'Data retrieved successfully',
-              success: true,
-            },
-            claim: data?.data,
-          }
+      // Append the claim data to the appointments in the query data
+      const newAppointmentsQueryData = validAppointmentQueries.map(([queryKey, queryData]) => {
+        const newQueryData = {
+          data: queryData?.data.map((appointment) => {
```
liztownd, line 50: From a performance perspective I wonder if there's a better way to do this, like maybe `queryData.data[id].travelPayClaim = {new claim object}` or something, rather than mapping over every appointment and checking the id. Not sure the performance boost would be worth it though. 🤔

VAMobile/src/api/travelPay/submitClaim.tsx
```diff
@@ -27,43 +27,41 @@ export const useSubmitTravelClaim = () => {
...
-        const newAppointment = { ...appointment }
+      // Filter out to only the queries where an appointment can submit a claim and that appointment is in the query data
+      const validAppointmentQueries = appointmentQueries.filter(([queryKey, queryData]) => {
+        const timeFrame = queryKey[1] as TimeFrameType
+        return (
+          [TimeFrameTypeConstants.PAST_THREE_MONTHS, TimeFrameTypeConstants.PAST_ALL_CURRENT_YEAR].includes(
+            timeFrame,
+          ) &&
+          !!queryData &&
+          queryData.data.some((appointment) => appointment.id === variables.appointmentID)
```
liztownd, line 43: We should only have one appointment that matches the ID passed in right? So maybe `find` instead of `some`, just to narrow it down? Or can the past 3 months & past all current year queries co-exist in the query cache (not sure exactly how that works)?

VAMobile/src/api/travelPay/submitClaim.tsx
liztownd: And now that I'm asking the other question about co-existing timeframes in the query cache I'm thinking maybe we do need to map it to catch multiple appointment lists?

juancstlm-a6: Wish we could but the `data` object is an array so we can't direct access it knowing the id, we have to loop :/ 

In the case where a user loads the one year and past 3 months and has those both cached there is duplication of data. So when an appointment is submitted under 3 months or one year the other time frame also needs updating

VAMobile/src/api/travelPay/submitClaim.tsx
juancstlm-a6: Yes, only one appointment should match the Id per time frame. It is possible the appointment can be in two time frames if the user loaded the past 3 months and past all year

VAMobile/src/api/travelPay/submitClaim.tsx
liztownd: Gotchya.

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/ClaimExamAppointment.tsx
juancstlm-a6: Updated. Most agree that given how many places we have to do this the approach you suggested Liz makes sense

VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/VideoVAAppointment.tsx
```diff
@@ -66,7 +66,7 @@ function VideoVAAppointment({
         <AppointmentReasonAndComment attributes={attributes} subType={subType} type={type} />
         <AppointmentMedicationWording subType={subType} type={type} />
         <AppointmentPersonalContactInfo attributes={attributes} subType={subType} />
-        <AppointmentTravelClaimDetails attributes={attributes} subType={subType} />
+        {featureEnabled('travelPaySMOC') && <AppointmentTravelClaimDetails attributes={attributes} subType={subType} />}
```
liztownd, line 69: Missed a feature flag in this one.

liztownd (approved):
🚀

PR #11550, opened 2025-08-26
830 - Adding New Overpayment Debt Endpoints

matt-guest-wilcore:
## Description of Change

As part of the overpay/copay feature we need to integrate new endpoints into the mobile app. For Overpayments, we need to implement these 2 new endpoints: `/v0/debts` and `/v0/debts/:id`.

**Note:** I can confirm `/v0/debts` is working but for some reason `/v0/debts/:id` has been returning a 500 error. It might be a temporary issue on the server-side but so far that endpoint hasn't yielded any results. Will keep testing on my end.

## Link to Issue

[830](https://github.com/department-of-veterans-affairs/va-mobile-feature-support/issues/830)

## Testing Requirements

- Login to a test user (81 or 90)
- Make a call to the `/v0/debts` and `/v0/debts/:id` endpoints by calling `useOverpaymentDebts()` and `useOverpaymentDebtById()`, respectively. When calling the first endpoint note down one of the `id`'s you see in the response and then use that id to pass as a param to `useOverpaymentDebtById("id goes here")`.

Acceptance Criteria

- [ ] Verify that endpoints return proper Overpayment Debt data

Test User(s)

Both of these users should have Overpay debts.

- [ ] Test User 81
- [ ] Test User 90

## Checklist for PR Submitter

### Administrative and documentation

- [x] Acceptance criteria is added on this PR or referenced from the attached issue

### Code testing

- [ ] Unit tests have been created or updated to cover this change
- [ ] End to end (Detox) tests have been created or updated to cover this change

### Code implementation

- [x] No secrets or API keys are checked present in the code

### New features

- [ ] Code is attached to a feature flag, or reason is given for no feature flag
- [ ] Design and UX has been approved by the code mobile team (documented on the PR or attached issue)

## Checklist for QA

- [ ] Tested on iOS
- [ ] Tested on Android


VAMobile/src/api/types/OverpaymentDebtData.ts
```diff
@@ -0,0 +1,33 @@
+export interface OverpaymentDebtHistoryEntry {
+  date?: string
+  letterCode?: string
+  description?: string
+}
+
+export interface OverpaymentDebtRecord {
+  id: string
```
oddballdave, line 8: See attached json captured from when the web calls the same endpoint.

- I don't see an `id` in the json
- The json has a `hasDependentDebts` value (do we need this?)
- The json has a `fiscalTransactionData` array with a lot more data
- The json has a `compositeDebtId` value

[debts.json](https://github.com/user-attachments/files/22013076/debts.json)

VAMobile/src/api/overpaymentDebts/getOverpaymentDebtById.tsx
oddballdave, line 1: I'm not aware of an endpoint where we can fetch a single debt by it's id??? If it doesn't exist, then this code can be removed.

oddballdave (commented):
One general comment is I'm wondering if we could rename modules/types from `OverpaymentDebt` to something simpler? Suggestions would be to use just `Overpayment(s)` or `Debts` everywhere. Not sure what makes sense, to name it by the API endpoint or a name similar to the way it's presented in the UI.

oddballdave (approved):

matthew-df (approved):

PR #11683, opened 2025-09-24
Removing myself from workflow and issue files so nothing fails after I'm gone

timwright12:
## Description of Change

Queuing up a PR to remove my user from files so workflows don't fail when my user gets removed

### DIFF
--- a/.github/ISSUE_TEMPLATE/detox_nightly_build_failure.md
+++ b/.github/ISSUE_TEMPLATE/detox_nightly_build_failure.md
@@ -3,7 +3,7 @@ name: Detox Nightly Build Failure Template
 about: Template for reporting a detox nightly build failure
 title: "Bug - Detox - Fix Overnight Failures"
 labels: bug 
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/detox_nightly_build_failure.md
```diff
@@ -3,7 +3,7 @@ name: Detox Nightly Build Failure Template
 about: Template for reporting a detox nightly build failure
 title: "Bug - Detox - Fix Overnight Failures"
 labels: bug 
-assignees: timwright12
+assignees: SKMOstudio
```
timwright12, line 6: @SKMOstudio this is a template for a ticket that gets created with Detox fails overnight. It's usually pretty quiet, but when Detox does fail it will create a ticket and assign it to you to look into. There isn't always something actionable, but if there is, just alert the team that owns the test (usually whoever last updated the file)

PR #12855, opened 2026-03-05
MHV SM - Add OH sync status alert for historic message loading

6StringCode:
## Description of Change

When a veteran's facility migrates to Oracle Health (OH), historic messages are synced asynchronously. During this period, the inbox may appear incomplete. This PR adds a warning alert to the Secure Messaging screen that displays when the OH sync status endpoint reports `syncComplete: false`, informing the veteran that their messages are still being loaded.

The sync status API call is **only made for users at pre-transitioned Oracle Health facilities** (`isUserAtPretransitionedOhFacility`). Non-OH users skip the call entirely and see no loading delay.

Changes:
- New `useOhSyncStatus` React Query hook calling `/v0/messaging/health/messages/oh_sync_status`
- `OhSyncStatusData` type added to `SecureMessagingData.ts` with proper nested data shape (`data.attributes.syncComplete`)
- Warning alert rendered in `SecureMessaging.tsx` when sync is incomplete
- **Sync status query gated to pre-transitioned OH users only** via `isUserAtPretransitionedOhFacility` from authorized services
- **Loading spinner only blocks for OH users** — non-OH users are not blocked by the sync status fetch
- **Sync complete state derived directly from query data** instead of `useState`/`useEffect`, avoiding stale state bugs
- Unit tests covering sync incomplete, sync complete, endpoint error, and **non-OH user** scenarios
- New translation strings for the alert title and body
- Demo mode support for the `oh_sync_status` endpoint via per-user mock JSON files
- Added e2e test in `OHMigrationMessages.e2e.ts` verifying the historic load alert appears when `syncComplete` is `false`

## Target Release Date

N/A

## Link to Issue

[#134304 - [SM mobile] - Display Message for Historic Message Load Timing on Mobile Platform
](https://github.com/department-of-veterans-affairs/va.gov-team/issues/134304)

## Screenshots/Video
Simulator Screenshot - iPhone 17 Pro - 2026-03-11 at 10 43 00

### Testing with Demo Mode Users

Two demo mode users are configured with different OH sync statuses to allow QA to test both states:

| Demo User | `syncComplete` | `isUserAtPretransitionedOhFacility` | Expected Behavior |
|-----------|---------------|-------------------------------------|-------------------|
| **Kimberly Washington** | `true` | `true` | No alert shown — normal messaging |
| **KimberlyForOHMigration** | `false` | `true` | Warning alert appears: "We're still adding some of your messages here" |

**How to test:**
1. Open the app in demo mode
[...]

PR #13030, opened 2026-03-24
Fix PushNotifications E2E by extracting OhSyncStatusAlert into conditional child component

narin:
## Description of Change

The PushNotifications E2E test has been consistently failing since PR #12855 merged on March 23.

### What we know

- A recurring native timer blocks Detox's idle detection during the PushNotifications E2E test, causing every interaction to hang for 6+ minutes and exceeding the 400s Jest timeout.
- This timer has always existed in the PushNotifications test — even passing runs (before PR #12855) showed it for ~45 seconds before it resolved. After PR #12855, it persists for 6+ minutes instead.
- The PushNotifications E2E is unique among tests: it launches the app with a simulated push notification (`userNotification`), which triggers iOS push notification registration and creates the native timer. Other E2E tests don't do this. This is related to Secure Messaging since the Push e2e tests is using an SM deep link as an example, which navigates to the same screen as the OH alert.
- We confirmed via git ancestry and E2E runs that removing `useOhSyncStatus` from `SecureMessaging.tsx` resolves the timeout ([passing run](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/23500324650)), and adding it back as a conditionally-mounted child component also resolves it ([passing run](https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/23501302973)).
- The default demo user has `isUserAtPretransitionedOhFacility: false`, so the hook's `enabled` condition is `false` and the query never fires — yet the QueryObserver is still created when the hook is called unconditionally in the parent component body.

### What we don't fully understand

The exact mechanism by which a disabled QueryObserver extends the native timer from ~45s to 6+ minutes isn't clear. It may be that the additional observer tips the app past a threshold where Detox can no longer detect idle state during the push notification registration flow. We're investigating this separately with enhanced Detox sync logging.

### The fix

Extract the `useOhSyncStatus` call into an inline `OhSyncStatusAlert` child component that only mounts when the user is a pre-transitioned OH facility user AND on the inbox tab. For all other users (including the default demo user in E2E), the component never mounts, so no QueryObserver is created. Behavior for OH users is unchanged.

`SecureMessaging.tsx` — Remove `useOhSyncStatus` from the parent component body and the associated loading gate. Add an inline `OhSyncStatusAlert` child component that owns the hook call and alert rendering, conditionally mounted only for pre-transitioned OH users.

## Target Release Date
N/A

## Code testing

- [ ] Unit tests have been created or updated to cover this change
- [x] End to end (Detox) tests have been added or updated as needed

## Pre-QA Artifacts

- [x] Visual artifacts not applicable to this PR (explain why below)

No visual changes — this is a refactor that moves existing UI into a child component with identical rendering behavior for OH users.

## Test Context for QA

**How does a user get here?**
Health tab → Messages (Secure Messaging inbox). The OH sync status alert only appears for users at pre-transitioned Oracle Health facilities.

**Feature Flags**
No feature flags involved.

**Risk Assessment:**
- [x] **Low** (UI polish, copy change, isolated component)

**What should QA pay extra attention to?**
Verify the OH sync status alert still renders correctly for OH migration test users (e.g., kimberlyForOHMigration demo user). For regular users, behavior should be unchanged.

**Test Review**
No test file changes. The existing PushNotifications E2E and SecureMessaging unit tests validate the fix.

**Test User(s)**
Default demo user (non-OH) for PushNotifications E2E. kimberlyForOHMigration demo user to verify OH sync alert still works.

---

## Checklist for QA

- [ ] [Shared Test Script](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/develop/VAMobile/QA/processes/testscript.md) executed (post results as a PR comment)
- [ ] Feature-specific verification based on QA Test Context above
- [ ] Tested on iOS
- [ ] Tested on Android


jmarchi (approved):

6StringCode (approved):
LGTM. Thanks for catching and flagging this.

digitalken (approved):
Approved.  Sync OhSyncStatusAlert displays when it is supposed to.

Image

## 6. Platform Parity

- [x] Tested on **iOS** (simulator or device)
- [x] Tested on **Android** (emulator or device)
- [x] Behavior is consistent across platforms (or platform-specific behavior is intentional and documented)

## 7. Data and State

- [x] Works with expected/valid test data (happy path)

## 8. Feature-Specific Verification
- [x] **Happy path**: Navigate to the feature using the path described in the PR and verify it works as described

digitalken (approved):

PR #13259, opened 2026-04-15
Remove unused actions: read from queue_builds permissions

narin:
## Summary

Fixes the v2.73.0 release build startup failure: https://github.com/department-of-veterans-affairs/va-mobile-app/actions/runs/24470502791

## Root cause

`release_build.yml` declares an explicit `permissions:` block with only `id-token: write` and `contents: write`. The reusable `queue_builds.yml` declared `actions: read`. A reusable workflow cannot request permissions the caller does not grant, so the run fails at startup (no jobs ran).

Other callers (`qa_build.yml`, `on_demand_build.yml`, `release_candidate_build.yml`) have no top-level `permissions:` block and inherit repo defaults, which is why this only breaks the release build.

## Fix

Remove `actions: read` from `queue_builds.yml`. It was unused: the `curl` to `/actions/runs` authenticates with a GitHub App token (`steps.generate_token.outputs.token`), not `GITHUB_TOKEN`.

## Next steps after merge

Per the Release Runbook (Re-releasing a failed build), the `v2.73.0` tag needs to be deleted and recreated pointing at the new commit so `release_build.yml` runs again. Retrying the failed run replays the old commit's workflow file and will not pick up this fix. Confirm no iOS submission is already pending in App Store Connect before re-tagging; re-tagging rebuilds both platforms.


narin:
Covered by https://github.com/department-of-veterans-affairs/va-mobile-app/pull/13260. Closing.

